PR #13, opened 2020-09-16
Hauki 103 integrate cypress

VilleMiekkoja:


cypress/integration/examples/actions.spec.js
```diff
@@ -0,0 +1,299 @@
+/// <reference types="cypress" />
```
minevala, line 1: What convention should we follow with the generated examples files? 🤔 
Should we keep them or replace them with a readme-link?

.gitignore
```diff
@@ -7,6 +7,10 @@
 
 # testing
 /coverage
+/cypress/videos/*
+!/cypress/videos/.gitkeep
```
minevala, line 11: I'm seeing this as deleted file after running tests 🤔

cypress/integration/examples/actions.spec.js
VilleMiekkoja: Yeah readme-link is probably better, gonna add one and delete these examples.

- [ ] Delete examples and provide link to them.

.gitignore
VilleMiekkoja: Should be fixed now

- [ ] Fix cypress deleting .gitkeep

minevala (approved):
Good job

PR #17, opened 2020-09-21
Fix problems with hds navigation rendering in production build

VilleMiekkoja:
Sub navigation bar rendering issue still remains. Fix for that is under it's way by hds team.

EDIT: Nvm, hds was fast and they already released a fix. It will be included in this pr


src/components/navigation/HaukiNavigation.css
```diff
@@ -1,3 +1,3 @@
-.navigation-header {
+.navigation-header.navigation-header.navigation-header {
```
VilleMiekkoja, line 1: Not optimal and pretty, can we find a cleaner fix?

minevala: Alternative options:
```css
 .navigation-header {
  --header-background-color: var(--primary-color) !important; // We need to force override with !important
}
```

```css
 header.navigation-header { // We need more specific selector to override the default hds-styles
  --header-background-color: var(--primary-color) !important; // We need to force override with !important
}
```

minevala: Would it make sense to make all of the customizations in one place? 
```css
:root {
  --primary-color: #6feed3;
  --header-background-color: var(--primary-color) !important;
}
```

minevala (commented):
Nice that the solution is on it's way  👍

src/components/navigation/HaukiNavigation.css
VilleMiekkoja: > Alternative options:
> 
> ```css
>  .navigation-header {
>   --header-background-color: var(--primary-color) !important; // We need to force override with !important
> }
> ```
> 
> ```css
>  header.navigation-header { // We need more specific selector to override the default hds-styles
>   --header-background-color: var(--primary-color) !important; // We need to force override with !important
> }
> ```

Those would work also, though not sure if they would be cleaner. mm..

minevala: I think the most cleanest would be to override them all in on place 🤔

VilleMiekkoja: Good point!

VilleMiekkoja: - [ ] Override variables in one place

minevala (approved):
Nice 👍

PR #19, opened 2020-09-23
Hauki 83 react axe

VilleMiekkoja:


cypress/support/index.js
```diff
@@ -1,5 +1,5 @@
 // ***********************************************************
```
minevala, line 1: Would you change the extensions of the other files from .js to .ts too?

cypress/integration/App.spec.ts
```diff
@@ -1,6 +1,14 @@
+/// <reference types="cypress-axe" />
+
 describe('Open aukiolot app', () => {
```
minevala, line 3: I get the following lint warning about this in my IDE: 
`TS1208: All files must be modules when the '--isolatedModules' flag is provide`

minevala (commented):
I commented on a couple of small things. It is nice that we also have an automated way to test accessibility 👍

cypress/integration/App.spec.ts
minevala: Näyttökuva 2020-9-23 kello 17 23 55

VilleMiekkoja: This should be fixed now

minevala (approved):
🥇

PR #20, opened 2020-09-24
Hauki 129 main content component

minevala:
Add main-component to align main content with the header content.

Näyttökuva 2020-9-24 kello 10 10 32

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "enzyme-adapter-react-16": "^1.15.4",
     "eslint-config-airbnb-typescript-prettier": "^3.1.0",
     "hds-core": "^0.13.2",
+    "hds-design-tokens": "^0.13.2",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -72,6 +73,7 @@
   "devDependencies": {
     "cypress": "^5.1.0",
     "eslint-plugin-jsx-a11y": "^6.3.1",
+    "node-sass": "^4.14.1",
```
VilleMiekkoja, line 76: Tulisko olla dependencien puolella?

VilleMiekkoja:
Looking good! 👍

package.json
minevala: I think node-sass is not included in the bundle like React or HDS. That is why I put it into devDependencies.

VilleMiekkoja (approved):

PR #21, opened 2020-09-28
Hauki 141 linter bug

minevala:
- Add separate type checking steps for both existing Typescript configurations, **tsconfig.json**, and **cypress/tsconfig.json**. 
The configs have contradictory rules, so it is not possible to combine them.
 (Eslint itself does not compile Typescript code hence does not check types.)
- Include files from both Typescript configurations for Eslint with **tsconfig.eslint.json**:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/parser/README.md#parseroptionsproject


VilleMiekkoja:
What about our public folder? Should we apply linting there also? There's only one js file there though. My IDE is showing a warning if I open the file:

image

VilleMiekkoja (approved):

minevala:
I added an override rule for js-files

PR #22, opened 2020-09-29
Master pipeline

VilleMiekkoja:
Initial master pipeline for the purpose of continuously releasing master branch to dev environment.

In its current form it does:

1. Runs tests
2. Builds and pushes docker image to dockerhub
3. Releases that image to dev env hosted by red hat openshift

Re-jira'd https://helsinkisolutionoffice.atlassian.net/browse/HAUKI-145 to run e2e tests automatically after successful release


nginx/nginx.conf
```diff
@@ -1,11 +1,47 @@
...
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    keepalive_timeout  65;
+
+    #gzip  on;
```
VilleMiekkoja, line 27: - [ ] Clean these comments and unnecessary empty lines away

scripts/env.sh
```diff
@@ -2,12 +2,8 @@
 
 set -e -o pipefail
 
-# Recreate config file
-rm -rf ./env-config.js
```
minevala, line 6: I think we still need remove the old one when we are running the script in the local env 🤔

release-master-to-test-environment.yml
```diff
@@ -0,0 +1,57 @@
+# Add steps that build, run tests, deploy, and more:
```
minevala, line 1: Do we need separate pipeline to make the test deploy? Or is there a way to extend the azure-pipeline.yml?

minevala (commented):
Nice 🔥

release-master-to-test-environment.yml
minevala: Or maybe we can group them somehow? by name? by using folder?

VilleMiekkoja: We discussed this and decided we'll put those pipeline files into a separate folder:

- [ ] Put azure pipeline files to a separate folder

minevala (approved):

PR #28, opened 2020-10-06
Hauki 154 tprek link

minevala:
- Add links to Target data model
- Add Link component
- Show Admin-link on target-page

Näyttökuva 2020-10-7 kello 16 37 40

### DIFF
--- a/src/common/utils/api/api.test.ts
+++ b/src/common/utils/api/api.test.ts
@@ -1,7 +1,7 @@
 /// <reference types="jest" />
 
 import axios from 'axios';
-import api, { Target } from './api';
+import api, { LinkTypes, Target } from './api';
 
 jest.mock('axios');
 const mockedAxios = axios as jest.Mocked<typeof axios>;
@@ -13,6 +13,8 @@ describe('apiRequest - getTarget', () => {
       name: 'Toimipiste A',
       address: 'Helsinki',
       description: '',
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      links: [{ link_type: LinkTypes.ADMIN, url: 'http://local.trek:3001' }],
     };
 
     mockedAxios.request.mockResolvedValue({ data: mockTarget });
--- a/src/common/utils/api/api.test.ts
+++ b/src/common/utils/api/api.test.ts
@@ -13,8 +13,8 @@ describe('apiRequest - getTarget', () => {
       name: 'Toimipiste A',
       address: 'Helsinki',
       description: '',
-      // eslint-disable-next-line @typescript-eslint/camelcase
       links: [
+        // eslint-disable-next-line @typescript-eslint/camelcase
         { link_type: SourceLinkTypes.ADMIN, url: 'http://local.trek:3001' },
       ],
     };
--- a/src/common/utils/api/api.test.ts
+++ b/src/common/utils/api/api.test.ts
@@ -14,7 +14,6 @@ describe('apiRequest - getTarget', () => {
       address: 'Helsinki',
       description: '',
       links: [
-        // eslint-disable-next-line @typescript-eslint/camelcase
         { link_type: SourceLinkTypes.ADMIN, url: 'http://local.trek:3001' },
       ],
     };
--- a/src/common/utils/api/api.test.ts
+++ b/src/common/utils/api/api.test.ts
@@ -1,7 +1,7 @@
 /// <reference types="jest" />
 
 import axios from 'axios';
-import api, { LinkTypes, Target } from './api';
+import api, { SourceLinkTypes, Target } from './api';
 
 jest.mock('axios');
 const mockedAxios = axios as jest.Mocked<typeof axios>;
@@ -14,7 +14,9 @@ describe('apiRequest - getTarget', () => {
       address: 'Helsinki',
       description: '',
       // eslint-disable-next-line @typescript-eslint/camelcase
-      links: [{ link_type: LinkTypes.ADMIN, url: 'http://local.trek:3001' }],
+      links: [
+        { link_type: SourceLinkTypes.ADMIN, url: 'http://local.trek:3001' },
+      ],
     };
 
     mockedAxios.request.mockResolvedValue({ data: mockTarget });
--- a/src/components/link/Link.tsx
+++ b/src/components/link/Link.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import './Link.scss';
+
+export default function Link({
+  href,
+  text,
+  external,
+}: {
+  href: string;
+  text: string;
+  external: boolean;
+}): JSX.Element {
+  return (
+    <a
+      href={href}
+      className="custom-link"
+      {...(external ? { rel: 'noopener noreferrer', target: '_blank' } : {})}>
+      {text}
+    </a>
+  );
+}
--- a/src/components/link/Link.tsx
+++ b/src/components/link/Link.tsx
@@ -7,7 +7,6 @@ export function Link({
 }: {
   href: string;
   text: string;
-  external: boolean;
 }): JSX.Element {
   return (
     <a href={href} className="custom-link">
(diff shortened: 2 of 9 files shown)


src/common/utils/api/api.test.ts
```diff
@@ -13,6 +13,10 @@ describe('apiRequest - getTarget', () => {
       name: 'Toimipiste A',
       address: 'Helsinki',
       description: '',
+      links: [
+        // eslint-disable-next-line @typescript-eslint/camelcase
```
minevala, line 17: This linter rule is a bit complex, not sure do we need this 🤔

src/components/link/Link.tsx
```diff
@@ -0,0 +1,21 @@
...
+}: {
+  href: string;
+  text: string;
+  external: boolean;
+}): JSX.Element {
+  return (
+    <a
+      href={href}
+      className="custom-link"
+      {...(external ? { rel: 'noopener noreferrer', target: '_blank' } : {})}>
```
VilleMiekkoja, line 17: Semisti kryptinen tämä rivi :D. Mietin saisko tätä yksinkertaistettua jotenkin? Ilmeisesti tässä halutaan erotella se, että avautuuko linkin sivu kokonaan uuteen tabiin, vai pysytäänkö sovelluksen sisällä? Vois olla luettavampi siten, että kumpikin attribuutti (rel ja target) olis omilla riveillään, vaikkakin siitä tulee hieman toistoa

src/components/link/Link.tsx
```diff
@@ -0,0 +1,21 @@
...
+import './Link.scss';
+
+export default function Link({
+  href,
+  text,
+  external,
+}: {
+  href: string;
+  text: string;
+  external: boolean;
```
VilleMiekkoja, line 11: Tän external propsin vaikutukset on ehkä hieman implisiittisiä. Mietin olisko joku toinen nimi kuvaavampi? Tyyliin: 'openToNewTab' tjsp.?

src/target/page/TargetPage.test.tsx
```diff
@@ -1,14 +1,16 @@
...
+import api, { Target, SourceLinkTypes } from '../../common/utils/api/api';
 import TargetPage from './TargetPage';
 
 const testTarget: Target = {
   id: 'tprek:1000',
   name: 'Test target',
   address: 'Main street, Helsinki',
   description: 'Test target description',
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  links: [{ link_type: SourceLinkTypes.ADMIN, url: 'http://local.trek:3001' }],
```
VilleMiekkoja, line 13: Dummie question: why we need link_type in snake_case?

src/target/page/TargetPage.tsx
```diff
@@ -1,12 +1,26 @@
...
+import Link from '../../components/link/Link';
 import './TargetPage.scss';
 
 const hasText = (str: string | null | undefined): boolean =>
   str !== undefined && str !== null && str !== '';
 
+const getTargetServiceAdminUrl = (
+  links: SourceLink[] = []
+): string | undefined => {
+  const adminLink = links.find(
```
VilleMiekkoja, line 18: Can we return the result of this find operation directly, or is it easier to create this const first and then return it in line 21?

src/target/page/TargetPage.tsx
```diff
@@ -28,13 +54,42 @@ const TargetDetailsSection = ({
...
+
+  if (!url) {
+    return null;
+  }
+
+  return (
+    <TargetSection id={id}>
+      <p>
+        Toimipisteeseen liittyvät tiedot kieliversioineen ovat lähtöisin
+        Toimipisterekisteristä. Jos haluat muuttaa tietoja, se on mahdollista
```
VilleMiekkoja, line 81: Would maybe simplify this sentence 'Jos haluat muuttaa tietoja, se on mahdollista Toimipisterekisterissä' -> 'Tietojen muuttaminen on mahdollista Toimipisterekisterissä'

VilleMiekkoja:
Good job! :fire:

src/common/utils/api/api.test.ts
VilleMiekkoja: Couple options come to my mind on how to resolve this situation:

1. We get the link_type returned to us as linkType
2. We map the link_type to linkType at the edge of our app. So that there's only one place where we need to specify this //eslint-disable-next-line rule
3. We create our own eslint rule, which allows snake_case, but only for variable named link_type: https://eslint.org/docs/developer-guide/working-with-rules

I think the 3rd option feels most cleanest, but might include most work?

src/components/link/Link.tsx
minevala: I can try to make it a bit more readable 👍

src/target/page/TargetPage.test.tsx
minevala: I thought that it could be as close as possible with the API response format without any unnecessary abstraction. 
But having said that, I could try some data conversion (api->client) and see how does it look..

src/components/link/Link.tsx
minevala: My idea behind this was that we could have different default behavior (and maybe design) for internal and external links.

src/common/utils/api/api.test.ts
minevala: I examined no.1, and it seems that most of the Helsinki APIs follow snakecase-convention. 
I'll test number 2 because it is the easiest. And if it fails, I'll go with number 3.

src/target/page/TargetPage.tsx
minevala: Let's see how does it work 👀

src/target/page/TargetPage.tsx
minevala: I just followed the design, but yes, it felt a bit complex for me too. I can iterate it a bit, and let's see how it goes.

src/components/link/Link.tsx
VilleMiekkoja: Yeah I understood that. I think it would be ok if that default behaviour could be easily derived from the api (either by explanation or variable name). But not sure how to do that. My fear is, that to understand what is happening, developers would have to dig in deeper to know what the effect of having external link is. Instead of just looking at the Link api (props), and seeing directly from there how the Link component is going to behave based on props.

src/target/page/TargetPage.test.tsx
VilleMiekkoja: I fear we end up in client codebase with 2 types of variable naming, which would lead to confusion as to which one to use hmm

src/components/link/Link.tsx
VilleMiekkoja: What if we create separate components: Link and ExternalLink? That would reduce the amount of conditions and emphasize more that we have two types of links hmm

src/common/utils/api/api.test.ts
VilleMiekkoja: After discussing and thinking about this more, it seems like we will have more variables that will be returned to us in snake_case format. Im worried there will be pain in any of the options; we would have to map all those variables, or to declare an eslint rule for all of those variables. Or then just allow different naming conventions for our variables. Im not sure which would be the option with the least pain? What do you think?

minevala: Yes, I'm thinking about that too after implementing one normalize function. I'll go with number 3

src/components/link/Link.tsx
minevala: I'll go with separate ExternalLink 👍

src/target/page/TargetPage.test.tsx
minevala: I think we can go with creating new linter rule or something

src/common/utils/api/api.test.ts
minevala: I added custom rules for naming-conventions

VilleMiekkoja (approved):

PR #29, opened 2020-10-08
Hauki 162 ui tweaking

minevala:
- Fix main's and footer's background-colors
- Set main's content line-height and adjust spacings. Fixes error notification icon-alignment and improves target's description's readability
- Preserve target's description's line-break-formatting to show lists in text correctly

Näyttökuva 2020-10-8 kello 15 18 06

### DIFF
--- a/src/target/page/TargetPage.tsx
+++ b/src/target/page/TargetPage.tsx
@@ -71,7 +71,7 @@ export default function TargetPage({ id }: { id: string }): JSX.Element {
     <>
       <TargetInfo target={target} />
       <TargetDetailsSection id="target-description" title="Perustiedot">
-        <p>
+        <p className="target-description-text">
           {hasText(target?.description)
             ? target?.description
             : 'Toimipisteellä ei ole kuvausta.'}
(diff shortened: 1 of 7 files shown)


src/target/page/TargetPage.tsx
```diff
@@ -9,7 +9,7 @@ const hasText = (str: string | null | undefined): boolean =>
 
 const TargetInfo = ({ target }: { target?: Target }): JSX.Element => (
   <>
-    <h1>{target?.name}</h1>
+    <h1 className="target-info-title">{target?.name}</h1>
```
VilleMiekkoja, line 12: What should we display here is name is missing? Could we add something like 'Nimeämätön kohde'? I know this is little offtopic regarding what this pr is about. This probably is not a priority, and also pretty rare condition and situation if the name would be missing.

VilleMiekkoja:
Well done! :star:

VilleMiekkoja (approved):

src/target/page/TargetPage.tsx
minevala: I can but a default text: "Toimipisteellä ei ole nimeä"

PR #31, opened 2020-10-13
For consideration: separate test & build stages in pipeline

vikoivun:
This is more of a proof of concept (E2E tests will probably not work).

Tests are now run in first stage and image build in second stage. This ensures that anything changed durings tests does not affect the builds. Previously the builds were dependent on things done in tests?

### DIFF
--- a/azure/master-to-test.pipeline.yml
+++ b/azure/master-to-test.pipeline.yml
@@ -18,57 +18,62 @@ pool:
 
 stages:
 - stage: Tests
-  steps:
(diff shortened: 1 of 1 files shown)


vikoivun:
Also this needs to be checked, as it seems that Stages are not guaranteed to be run on a separate agent.

minevala:
I'm not sure if this PR is valid anymore since we are not using the current CI-pipelines in the future.

PR #32, opened 2020-10-13
Hauki 171 fix docker build

minevala:
- Add dockerignore to prevent copying of unnecessary files
- Fix issue with node-sass build


Dockerfile
```diff
@@ -24,9 +24,6 @@ USER appuser
 COPY --chown=appuser:appuser package.json yarn.lock /app/
 RUN yarn && yarn cache clean --force
 
-USER root
```
VilleMiekkoja, line 27: Was this section not needed?

README.md
```diff
@@ -58,7 +58,7 @@ docker build . -t hauki-admin-ui
 
 ### How to run the docker image
 ```bash
-docker run -p 3000:8000 -e API_URL=<api-url-here> USE_AXE=<true|false> hauki-admin-ui
+docker run -p 3000:8000 -e API_URL=<api-url-here> -e USE_AXE=<true|false> hauki-admin-ui
```
VilleMiekkoja, line 61: Good find :+1:

Dockerfile
minevala: It is not needed, the build works without it.

PR #33, opened 2020-10-16
Hauki 112 auth params to header

minevala:
- Add utility to read and write values from/to localStorage
- Add auth-context 
- Show username in header

Näyttökuva 2020-10-19 kello 16 34 56


src/auth/auth-context.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+export const convertParamsToTokens = (item: ParsedUrlQuery): Tokens =>
+  (item as unknown) as Tokens;
+
+export type AuthContextProps = {
+  tokens: Tokens;
+};
+
+export const AuthContext = createContext<Partial<AuthContextProps>>({});
```
VilleMiekkoja, line 52: Typescript-noob kysymys: Mikä tällä rivillä tyypitetään, tuon kutsun paluuarvo vaih?

src/common/utils/storage/storage.ts
```diff
@@ -0,0 +1,19 @@
+export function getStorageItem<T>(key: string): T | undefined {
```
VilleMiekkoja, line 1: If only place where the functions of this file are applied is in auth-context.ts file, then I think it might be too early to refactor this to its own module. It makes reading little bit harder imo.

src/auth/auth-context.ts
minevala: ääh mulla on väärässä kohdin tyypitystä

src/App.tsx
```diff
@@ -1,40 +1,65 @@
...
+} from './auth/auth-context';
 import Main from './components/main/Main';
 import NavigationAndFooterWrapper from './components/navigation-and-footer-wrapper/NavigationAndFooterWrapper';
 import './App.scss';
 import TargetPage from './target/page/TargetPage';
 
 export default function App(): JSX.Element {
+  const existingTokens: Tokens | undefined = getTokens();
+
+  const parameterTokens: ParsedUrlQuery = querystring.parse(
```
VilleMiekkoja, line 26: Could we apply term authParams here? Would better fit with the functions `isValidAuthParams` and `converParamsToTokens` :)

src/App.tsx
```diff
@@ -1,40 +1,65 @@
...
+  isValidAuthParams,
+  AuthContext,
+} from './auth/auth-context';
 import Main from './components/main/Main';
 import NavigationAndFooterWrapper from './components/navigation-and-footer-wrapper/NavigationAndFooterWrapper';
 import './App.scss';
 import TargetPage from './target/page/TargetPage';
 
 export default function App(): JSX.Element {
+  const existingTokens: Tokens | undefined = getTokens();
```
VilleMiekkoja, line 24: I wonder if the term _Token_ could be unnecessarily abstract for us? Would it be more readable with a prefix _auth_ in it to indicate what those tokens are? In other words, tokens obtained from authentication. :thinking:

VilleMiekkoja:
Good job! :v:

src/App.tsx
minevala: Do you mean the name of the type and variables? I'll test how does it look

src/auth/auth-context.ts
minevala: I'll fix this

src/components/navigation/HaukiNavigation.tsx
```diff
@@ -51,7 +54,7 @@ export default function HaukiNavigation(): JSX.Element {
           <Navigation.Item
             as="button"
             type="button"
-            onClick={(): void => setAuthenticated(false)}
+            onClick={(): void => console.log('plaa')}
```
VilleMiekkoja, line 57: Are we confident that hits console.log will not be forgotten here? :D.

minevala: Nope 😁 
I'll remove those

src/auth/auth-context.ts
minevala: fixed

src/common/utils/storage/storage.ts
minevala: Refactored to auth-context

src/App.tsx
minevala: Renamed

src/App.tsx
minevala: Replaced tokens with authTokens

src/components/navigation/HaukiNavigation.tsx
minevala: Removed the whole signout-button

VilleMiekkoja (approved):

PR #41, opened 2020-10-22
Hauki 170 check auth api

minevala:
- Add service to test credentials
- Validate the credentials with api request:
   a. store them and set them into auth context
   b. log.error and remove old ones if they exist
- Use auth-query-string as env variable in authenticated user e2e-tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ See the section about [running tests](https://facebook.github.io/create-react-ap
 
 **Prequisites:**
 All e2e-tests require an environment variable `CYPRESS_AUTH_QUERY_PARAMETERS`.
+Parameters can be generated for local tests by using the Hauki API's query-generating view (only available locally).
 
 Runs cypress (e2e) tests locally.
 
--- a/README.md
+++ b/README.md
@@ -21,10 +21,6 @@ See the section about [running tests](https://facebook.github.io/create-react-ap
 
 ### `yarn test-cypress`
 
-**Prequisites:**
-All e2e-tests require an environment variable `CYPRESS_AUTH_QUERY_PARAMETERS`.
-Parameters can be generated for local tests by using the Hauki API's query-generating view (only available locally).
-
 Runs cypress (e2e) tests locally.
 
 ### `yarn build`
--- a/README.md
+++ b/README.md
@@ -21,6 +21,9 @@ See the section about [running tests](https://facebook.github.io/create-react-ap
 
 ### `yarn test-cypress`
 
+**Prequisites:**
+All e2e-tests require an environment variable `CYPRESS_AUTH_QUERY_PARAMETERS`.
+
 Runs cypress (e2e) tests locally.
 
 ### `yarn build`
--- a/cypress/integration/User.spec.ts
+++ b/cypress/integration/User.spec.ts
@@ -1,14 +1,18 @@
 /// <reference types="cypress" />
 /// <reference types="node"/>
-import querystring from 'querystring';
+/// <reference path="../index.d.ts" />
+import * as querystring from 'querystring';
 
-const authObject = {
-  username: 'admin@hel.fi',
-  organization: 'dc7d39db-b35a-4612-a921-1b7b24b0baa3',
-  resource: 'tprek:0010',
-  created_at: '2020-10-14',
-  valid_until: '2020-10-14',
-  signature: '50A2A31D50F776D23A87F06708934C43CC58B3E44F851018675266393A457828',
+// TODO: later on username will come from env variables.
+const parseUserName = (queryParams: string): string | undefined => {
+  const userNameParameters: string | string[] = querystring.parse(queryParams)
+    ?.username;
+  const userNameStr: string =
+    typeof userNameParameters === 'string'
+      ? userNameParameters
+      : userNameParameters[0];
+
+  return userNameStr ? decodeURIComponent(userNameStr) : undefined;
 };
 
 describe('Unauthenticated user', () => {
@@ -28,6 +32,8 @@ describe('Authenticated user', () => {
 
   it('Has username in header', () => {
     cy.get('header').first().should('not.contain', 'Kirjaudu');
-    cy.get('header').first().should('contain', authObject.username);
+    cy.get('header')
+      .first()
+      .should('contain', parseUserName(Cypress.env('AUTH_QUERY_PARAMETERS')));
   });
 });
--- a/cypress/integration/User.spec.ts
+++ b/cypress/integration/User.spec.ts
@@ -1,19 +1,6 @@
 /// <reference types="cypress" />
-/// <reference types="node"/>
 /// <reference path="../index.d.ts" />
-import * as querystring from 'querystring';
-
-// TODO: later on username will come from env variables.
-const parseUserName = (queryParams: string): string | undefined => {
-  const userNameParameters: string | string[] = querystring.parse(queryParams)
-    ?.username;
-  const userNameStr: string =
-    typeof userNameParameters === 'string'
-      ? userNameParameters
-      : userNameParameters[0];
-
-  return userNameStr ? decodeURIComponent(userNameStr) : undefined;
-};
+import { parseAuthParams } from '../../src/auth/auth-context';
 
 describe('Unauthenticated user', () => {
   beforeEach(() => {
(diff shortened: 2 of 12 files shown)


src/App.tsx
```diff
@@ -9,58 +9,87 @@ import 'hds-core';
...
-    authTokensFromQuery || existingAuthTokens;
+    const authTokensFromQueryOrStore: OptionalAuthTokens =
+      authTokensFromQuery || storedAuthTokens;
 
-  useEffect(() => {
-    if (authTokens) {
-      setTokens(authTokens);
+    if (authTokensFromQueryOrStore) {
+      api
+        .testAuthCredentials(authTokensFromQueryOrStore)
```
VilleMiekkoja, line 48: Why do we need to separately test these tokens on page load? Is testing not done while we retrieve the info of the target(resource)?

VilleMiekkoja: Discussed about this on slack. We need to make a separate auth check because most of the GET apis are public, and accessible even without any auth.

src/auth/auth-context.ts
```diff
@@ -41,6 +39,15 @@ export const setTokens = (
   }
 };
 
+export const removeTokens = (): void => {
+  try {
+    window.localStorage.removeItem(tokenStorageKey);
+  } catch (error) {
```
VilleMiekkoja, line 45: I wonder if we would wish to not catch this error here? Or catch it but throw another error? Because the user of removeTokens function would probably like to know if the operation succeeded or not?

src/common/utils/api/api.ts
```diff
@@ -63,7 +76,18 @@ export interface Target {
   links: [SourceLink];
 }
 
+export interface AuthTestResponse {
```
VilleMiekkoja, line 79: Do we need to export this?

cypress/integration/User.spec.ts
```diff
@@ -1,19 +1,23 @@
...
+import * as querystring from 'querystring';
 
-const authObject = {
-  username: 'admin@hel.fi',
-  organization: 'dc7d39db-b35a-4612-a921-1b7b24b0baa3',
-  resource: 'tprek:0010',
-  created_at: '2020-10-14',
-  valid_until: '2020-10-14',
-  signature: '50A2A31D50F776D23A87F06708934C43CC58B3E44F851018675266393A457828',
+// TODO: later on username will come from env variables.
```
VilleMiekkoja, line 6: Hmm.. is this comment obsolete?

src/auth/auth-context.ts
minevala: 🤔 Should I actually detect if the user even has invalid tokens in the store? Then we would not need the try-catch-block

cypress/support/commands.ts
```diff
@@ -23,3 +23,13 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add(
+  'visitTargetPage',
```
VilleMiekkoja, line 28: Random idea that came to my mind while reading this: would it be more readable if we refactor this command to 2 different commands:

1. visitTargetPageAsAuthenticatedUser
2. visitTargetPageAsUnauthenticatedUser

:thinking:

cypress/integration/User.spec.ts
```diff
@@ -23,13 +27,13 @@ describe('Unauthenticated user', () => {
...
-    );
+    cy.visitTargetPage(Cypress.env('target-id'));
   });
 
   it('Has username in header', () => {
     cy.get('header').first().should('not.contain', 'Kirjaudu');
-    cy.get('header').first().should('contain', authObject.username);
+    cy.get('header')
+      .first()
+      .should('contain', parseUserName(Cypress.env('AUTH_QUERY_PARAMETERS')));
```
VilleMiekkoja, line 37: Do we need to parse the user name in this? Could we just hard-code the value here? It feels like it adds a possibility for an error. If our test function `parseUserName` works wrong, our test might fail, even our application code would work ok. Just not sure if I like that our tests depends heavily on our test function itself. Could we somehow apply the exact same function that our react-router is using? :D Probably not. :thinking:

src/auth/auth-context.ts
minevala: Like this maybe: 
```suggestion
  const saveAuthTokes = (tokens: OptionalAuthTokens) => {
    storeTokens(tokens);
    setAuthTokens(tokens);
    setLoading(false);
  };

  const onError = (e) => {
    // eslint-disable-next-line no-console
    console.error(`Authentication failed: ${e.message}`);
    setAuthTokens(undefined);
    setLoading(false);
  };

  useEffect(() => {
    const queryParams: ParsedUrlQuery = querystring.parse(
      window.location.search.replace('?', '')
    );
    const authTokensFromQuery: OptionalAuthTokens = pickAuthParams(queryParams);
    const storedAuthTokens: OptionalAuthTokens = getTokens();

    if (storedAuthTokens) {
      removeTokens();
    }

    if (authTokensFromQuery) {
      api
        .testAuthCredentials(authTokensFromQuery)
        .then(() => saveAuthTokes(authTokensFromQuery))
        .catch(onError);
    } else if (storedAuthTokens) {
      api
        .testAuthCredentials(storedAuthTokens)
        .then(() => saveAuthTokes(storedAuthTokens))
        .catch(onError);
    } else {
      setAuthTokens(undefined);
      setLoading(false);
    }
  }, []);

```

VilleMiekkoja: That could work. I just wonder if in theory it would be possible for something to remove the item after successfull check but before removal? So in theory it could always throw :thinking:

cypress/support/commands.ts
minevala: Yes, I also thought about this 👍 
Will fix

cypress/integration/User.spec.ts
minevala: Good points. I'll think of something else 👍

src/common/utils/api/api.ts
minevala: I'll check

VilleMiekkoja:
Looking good, well done! 🥇

src/auth/auth-context.ts
minevala: Yes, I will change that the removeTokens will always throw

cypress/integration/User.spec.ts
minevala: It might be 😁

cypress/integration/User.spec.ts
VilleMiekkoja: Could one solution be that we use data-* attributes that cypress recommends? https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

In short, instead of checking match for the actual name, we would check the existence of an element that should only appear after successful authentication. Or we could check that it needs to have some content in it. The use of data-* attributes feels most clean solution though

minevala: Hmm.. I can check if it is possible

README.md
```diff
@@ -21,6 +21,9 @@ See the section about [running tests](https://facebook.github.io/create-react-ap
 
 ### `yarn test-cypress`
 
+**Prequisites:**
+All e2e-tests require an environment variable `CYPRESS_AUTH_QUERY_PARAMETERS`.
```
VilleMiekkoja, line 25: Do we have a need for a package.json script for running cypress locally without the need of manually setting the environment variable for local tests?

minevala: Can you elaborate? I do not understand the questions 🤔 😁

VilleMiekkoja: "All e2e-tests require an environment variable `CYPRESS_AUTH_QUERY_PARAMETERS`." How to provide this environment variable if you wish to run cypress tests locally?

cypress/integration/User.spec.ts
```diff
@@ -23,13 +12,12 @@ describe('Unauthenticated user', () => {
...
 describe('Authenticated user', () => {
   beforeEach(() => {
-    cy.visit(
-      `/target/${Cypress.env('target-id')}?${querystring.stringify(authObject)}`
-    );
+    cy.visitTargetPageAsAuthenticatedUser(Cypress.env('target-id'));
   });
 
   it('Has username in header', () => {
+    cy.get('header', { timeout: 5000 }).should('be.visible');
```
VilleMiekkoja, line 19: I think the header might be there even when page is loading. If so, then it would be better to check element of the target page, like done in App.spec.ts

minevala: It has to be some element that we have control of. I'll use the same element as in App.spec

minevala: Fixed

VilleMiekkoja (approved):

PR #43, opened 2020-10-26
HAUKI-189 Fix content horizontal alignment

VilleMiekkoja:
image

### DIFF
--- a/src/components/main/Main.scss
+++ b/src/components/main/Main.scss
@@ -25,8 +25,6 @@
 @media screen and (min-width: $breakpoint-m) {
   .main-container {
     padding: $spacing-m;
-    padding-left: $spacing-m;
-    padding-right: $spacing-m;
   }
 }
 
--- a/src/components/main/Main.scss
+++ b/src/components/main/Main.scss
@@ -8,10 +8,16 @@
   background-color: var(--hauki-content-background-color);
 }
 
+.main {
+  display: flex;
+  justify-content: center;
+}
+
 .main-container {
   box-sizing: border-box;
   line-height: var(--lineheight-l);
-  margin: 0 auto;
+  margin-top: 0;
+  margin-bottom: 0;
   padding-top: $spacing-s;
   padding-bottom: $spacing-s;
   padding-left: 0;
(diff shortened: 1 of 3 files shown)


src/components/navigation/HaukiNavigation.scss
```diff
@@ -0,0 +1,8 @@
+.navigation-header {
+  nav {
+    a {
+      padding-left: 0 !important;
```
VilleMiekkoja, line 4: Had to apply this dirty hack; HDS nav component has a bug that prevents a more elegant fix. They promised to fix it soon.

src/components/main/Main.scss
```diff
@@ -22,9 +31,19 @@
...
 
-@media screen and (min-width: $breakpoint-m) {
+@media screen and (max-width: $breakpoint-xl) {
+  .main-container {
+    margin-left: $spacing-m;
+    margin-right: $spacing-m;
+  }
+}
+
+@media screen and (max-width: 840px) {
```
minevala, line 41: I changed the last lines to this (compared to master) and it seemed to fix some of the problems: 
```suggestion
@media screen and (min-width: $breakpoint-m) {
  .main-container {
    padding: $spacing-m;
    padding-left: $spacing-m;
    padding-right: $spacing-m;
  }
}

@media screen and (min-width: $breakpoint-xl) {
  .main-container {
    max-width: $container-width-xl;
    padding-left: 0;
    padding-right: 0;
  }
}
```

There might be some breakpoint that I have missed tough 🤔

VilleMiekkoja: This seems like a more elegant solution, let's use it instead.

src/components/main/Main.scss
```diff
@@ -25,11 +25,15 @@
 @media screen and (min-width: $breakpoint-m) {
   .main-container {
     padding: $spacing-m;
+    padding-left: $spacing-m;
+    padding-right: $spacing-m;
```
minevala, line 29: These two declarations are not needed anymore because the above shorthand. 
Maybe some copy-paste thingy from my suggestion 😬 😁

VilleMiekkoja: True, for both :D

minevala (approved):
Good improvement 👍

PR #44, opened 2020-10-29
Hauki 202 auth redirect

minevala:
- User with invalid parameters is redirected to the front-page (parameters are checked by using api endpoint)
- Add helper-scripts to simplify local development:
    - yarn start opens the application with required query-parameters


VilleMiekkoja:
Excellent! :star:

VilleMiekkoja (approved):

scripts/start-local-app.js
```diff
@@ -0,0 +1,14 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const childProcess = require('child_process');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const openBrowser = require('react-dev-utils/openBrowser');
+
+const [, , url] = process.argv;
+
+const queryParams = childProcess.execFileSync('node', [
```
VilleMiekkoja, line 8: Why do we need the child process?

src/App.tsx
```diff
@@ -85,8 +86,8 @@ export default function App(): JSX.Element {
                   <Route exact path="/">
                     <h1>Etusivu</h1>
                   </Route>
-                  <Route
-                    id="resource-route"
+                  <PrivateRoute
+                    id="target-route"
```
VilleMiekkoja, line 90: Will this change resource back to target on purpose? Maybe we need to first rebase this to master?

scripts/generate-auth-params.js
```diff
@@ -0,0 +1,37 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const crypto = require('crypto');
+
+const key =
+  process.env.APP_KEY ||
+  'd94c278085e49e807b67a58b8fcb92239b3fe881252f1938e3e0ca09611684b3790df3a22c08a32959309fdfcf074b4204c2f84a1ef03025621c4cf23a5521f24299ba857956a1fcf1b61715cb8154baeedba4e0409e636da9e91d7a618a4403e748c3a2';
+
+const username = process.env.APP_USER || 'admin@hel.fi';
+const resource = process.env.APP_RESOURCE;
```
VilleMiekkoja, line 9: How are these env variables meant to be used? Should we add something about them in readme? Also should we default to some resource and organization? If these are meant to be user configured by his own, should we name these with hauki-prefix to separate them from other possible environment variables used by other apps?

scripts/start-local-app.js
```diff
@@ -0,0 +1,14 @@
...
+  './scripts/generate-auth-params.js',
+]);
+
+const fullUrl = `${url}?${queryParams}`;
+process.env.BROWSER = ''; // We need to unset the BROWSER so the openBrowser won't call this script again. This was set in package.json. Now it will use the updated url to open the browser.
```
VilleMiekkoja, line 13: Where do we need/use this BROWSER env variable?

src/App.tsx
minevala: I'll fix that one

scripts/generate-auth-params.js
minevala: Hmm.., I do not think we will have other apps envs in our repo 🤔 
But I think we can use the HAUKI-prefix

scripts/start-local-app.js
minevala: In the package.json: https://github.com/City-of-Helsinki/hauki-admin-ui/blob/HAUKI-202-auth-redirect/package.json#L37
It is the way to intervene in how the browser is opened.

scripts/start-local-app.js
minevala: It is another node script that doesn't return a function or similar. It writes the result to stdout.

VilleMiekkoja (approved):

PR #46, opened 2020-11-05
Pass auth-parameters in every request

minevala:
~~Is blocked by: #47~~
Add request parameters and signature in every api request

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,11 +27,6 @@ export default function App(): JSX.Element {
   const [isLoading, setLoading] = useState<boolean>(true);
   const [authTokens, setAuthTokens] = useState<OptionalAuthTokens>();
 
-  const saveAuthTokes = (tokens: OptionalAuthTokens): void => {
-    storeTokens(tokens);
-    setAuthTokens(tokens);
-  };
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -43,28 +38,24 @@ export default function App(): JSX.Element {
...
-    if (authTokensFromQuery) {
-      api
-        .testAuthCredentials(authTokensFromQuery)
-        .then(() => {
-          saveAuthTokes(authTokensFromQuery);
-          setLoading(false);
-        })
-        .catch((e: Error) => onAuthFail(e.message));
-    } else if (storedAuthTokens) {
+    const storedAuthTokens: OptionalAuthTokens = getTokens();
```
VilleMiekkoja, line 46: Do we want to always get the tokens from localstorage, even if we already have access to the tokens from query above? I guess it makes the code simpler, but also now we need to always access (get) local storage even we would not need to. But I guess it is a good way to test that the _storeTokens_ part worked?

VilleMiekkoja:
Great work!

VilleMiekkoja (approved):

PR #47, opened 2020-11-05
Bugfix: include all auth params

minevala:
Add also organization and resource as mandatory parameters.


VilleMiekkoja (approved):
Good improvement!

PR #52, opened 2020-11-16
Hauki 226 upgrade hds libraries

minevala:
Upgraded HDS-libraries from 0.13.2 -> ~~0.16.0~~ 0.16.1
Breaking changes, [in release 0.15.0](https://github.com/City-of-Helsinki/helsinki-design-system/releases/tag/v0.15.0):
- colors
- navigation and language-selector

kuva2323


cypress/support/commands.ts
```diff
@@ -37,3 +37,15 @@ Cypress.Commands.add(
     cy.visit(`/resource/${resourceId}?${Cypress.env('auth-query-parameters')}`);
   }
 );
+
+Cypress.on('uncaught:exception', (err: Error): boolean | Error => {
+  if (
+    /ResizeObserver loop completed with undelivered notifications/.test(
```
VilleMiekkoja, line 43: This kind of feels like a hack, hmm... Do I understand this correct: the new hds library (I assume this error has something to do with them?) will throw an error with this message during our cypress tests? Does this happen also during our normal application usage? Should this be something we should report back to them?

src/resource/page/ResourcePage.test.tsx
```diff
@@ -27,6 +28,21 @@ const testResource: Resource = {
   },
 };
 
+const testDatePeriod: DatePeriod = {
+  id: '1',
```
VilleMiekkoja, line 32: Small thing: If I remember correct, the id is also type number?

src/resource/page/ResourcePage.test.tsx
```diff
@@ -27,6 +28,21 @@ const testResource: Resource = {
...
+  created: '20-16-2020',
+  modified: '20-16-2020',
+  is_removed: false,
+  name: { fi: '', sv: '', en: '' },
+  description: { fi: '', sv: '', en: '' },
+  start_date: '',
+  end_date: '20-11-2020',
```
VilleMiekkoja, line 39: Not sure if this date format was supported? I know that atleast this one is: yyyy-MM-dd. Though this is just test data so not as important for it to be accurate. Maybe if its applied as documentation how things work it might be of some value to use correct format. Though maybe api supports multiple formats?

src/resource/page/ResourcePage.test.tsx
```diff
@@ -75,6 +95,10 @@ describe(`<ResourcePage />`, () => {
       .spyOn(api, 'getResource')
       .mockImplementation(() => Promise.resolve(testResource));
 
+    jest
+      .spyOn(api, 'getDatePeriod')
+      .mockImplementation(() => Promise.resolve([testDatePeriod]));
+
     const resourcePage = mount(<ResourcePage id={testResource.id} />);
```
VilleMiekkoja, line 102: Actually fixed this in my current pr: the id here is not the same id that our UI applies in URL. Look here: http://hauki-test.oc.hel.ninja/v1/resource/tprek:8215/ According to that the id of that resource is "id": 1186,

VilleMiekkoja (approved):

cypress/support/commands.ts
minevala: Yes, and it is the new languageSelector what is causing this

minevala: I commented this in the HDS Slack

src/resource/page/ResourcePage.test.tsx
minevala: I'll fix it

src/resource/page/ResourcePage.test.tsx
minevala: I'll change these

src/resource/page/ResourcePage.test.tsx
minevala: Ok, I noticed that one so it is the "real" id 😁

minevala: I'll fix

VilleMiekkoja (approved):

cypress/support/commands.ts
minevala: https://helsinkisolutionoffice.atlassian.net/browse/HDS-526

minevala: And no, it does not affect normal application usage. There is no errors in console or anything. It happens when running the Cypress tests. The tests will fail without ignoring the exception.

PR #53, opened 2020-11-17
HAUKI-116 Page for user to create new opening period with basic info

VilleMiekkoja:
image

- Will do unit tests for the new page in a separate pull request.
- Datepicker component and its subcomponents are not unit tested here. I suggest we apply HDS datepicker once it is ready, and then we can rely their unit testings.
- Naming of datepicker class names was left intact from original; I was too lazy to change every classname to match our kebab-style. Since we're probably going to switch to HDS datepicker, I kind of thought it might not be worth the effort. Though it might not be as big of a task I thought so let's see.
- Proper form validation and visualisations of validation errors was also excluded from this. I will create a task in jira for it.


VilleMiekkoja:
Will do unit tests in a separate pull request

src/components/datepicker/Datepicker.scss
```diff
@@ -0,0 +1,176 @@
+@import '~hds-core/lib/components/text-input/text-input.css';
+
+.datepickerWrapper {
```
VilleMiekkoja, line 3: I left the naming of these datepicker classnames intact with the original. Should we change it to our style (kebab-style)? I think we should apply the HDS datepicker once it is ready, so I kind of thought it might not be worth the effort. Then again, it might take a long time for HDS datepicker to be ready, and having 2 different styles of classnames in the codebase is not optimal, and might be confusing 🤔

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+    Error | undefined
+  >(undefined);
+  const [isLoading, setLoading] = useState<boolean>(true);
+  const [periodBeginDate, setPeriodBeginDate] = useState<Date | null>(null);
+  const [periodEndDate, setPeriodEndDate] = useState<Date | null>(null);
+  const [submitStatus, setSubmitStatus] = useState<SubmitStatus>('init');
+
+  const onSubmit = async (data: Inputs): Promise<void> => {
+    try {
+      if (!resource) {
```
VilleMiekkoja, line 48: This one probably requires a better solution :sweat_smile:.

Problem here is:

If I remove the `if (!resource)` clause, then eslint will complain about row 52 that resource could be null/undefined:

image

I could try to solve it with inserting postfix '?' like this: `resource: resource?.id`. But then eslint will complain:

image

I think the problem is that `api.postDatePeriod` requires object that complies with DatePeriod object. And DatePeriod object always has value for resource. We could fix it so that we make it optional in the DatePeriod type definition. But Im not sure if that would give the correct signal. Because DatePeriod as an object got from database always has a resource related to it. It's like something it always should have. But since we open this page we do not yet necessarily know the resource id, because we load the resource info during the page load. So in the code it actually gets a value undefined in row 37 (default value is undefined). Maybe there's some workaround for this rather than throwing a new error?

We could set the resource field of DatePeriod type definition to optional, but I feel it would give wrong information about the object, since in practice it should always have a resource field value other than undefined.
Maybe there's some easy typescript way to handle this that I just missed because of typescript-noobness :sweat_smile:

src/components/datepicker/Datepicker.scss
minevala: This is fine since it is extracted to its own module 👍

src/common/utils/api/api.ts
```diff
@@ -86,8 +100,22 @@ async function apiGet<T>({ path, parameters = {} }: GetParameters): Promise<T> {
...
+    },
+    method: 'post',
+    data,
+    validateStatus(status) {
+      return status < 300;
+    },
+  });
+}
+
 export interface Resource {
```
minevala, line 117: Could this be moved to the under lib/types?

src/common/utils/api/api.ts
```diff
@@ -122,14 +150,20 @@ export default {
   getResource: (id: string): Promise<Resource> =>
     apiGet<Resource>({ path: `${resourceBasePath}/${id}` }),
 
-  getDatePeriod: (resourceId: string): Promise<DatePeriod[]> =>
```
minevala, line 125: I think the resource id can also be in format "tprek:8100" 🤔

src/opening-period/page/CreateNewOpeningPeriodPage.scss
```diff
@@ -0,0 +1,122 @@
...
+
+  &:focus {
+    border-color: var(--color-coat-of-arms-blue) !important;
+    outline-color: var(--color-coat-of-arms-blue);
+    color: var(--color-coat-of-arms-blue) !important;
+  }
+}
+
+.disabled-publish-button.disabled-publish-button {
+  background-color: #f2f2f2 !important;
```
VilleMiekkoja, line 119: Forgot to extract these as variables, will do that:

- [ ] Extract these colors as variables

src/opening-period/page/CreateNewOpeningPeriodPage.scss
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  &:focus {
+    border-color: var(--color-coat-of-arms-blue) !important;
+    outline-color: var(--color-coat-of-arms-blue);
+    color: var(--color-coat-of-arms-blue) !important;
+  }
+}
+
+.disabled-publish-button.disabled-publish-button {
```
VilleMiekkoja, line 118: It seems I've actually forgot to implement this disabled button to the component. 😅  Will add it to this pr:

- [ ] Implement disabled button

src/components/datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,449 @@
...
+import TimesList from './TimesList';
+
+function generateUniqueId(prefix = 'datepicker-id'): string {
+  return `${prefix}-${uniqueId()}`;
+}
+
+const dateRegex = /^\d{2}\.\d{2}\.\d{4}$/;
+const datetimeRegex = /^\d{2}\.\d{2}\.\d{4} \d{2}:\d{2}$/;
+const dateFormat = 'dd.MM.yyyy';
+const datetimeFormat = `${dateFormat} HH:mm`;
```
minevala, line 29: It would be nice if these two were in some centralized location.
Maybe under common/utils/date-time?

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+          en: null,
+        },
+        start_date: formatDate(
+          parse(data.openingPeriodBeginDate, 'dd.MM.yyyy', new Date()),
+          'yyyy-MM-dd'
+        ),
+        end_date: formatDate(
+          parse(data.openingPeriodEndDate, 'dd.MM.yyyy', new Date()),
+          'yyyy-MM-dd'
+        ),
```
minevala, line 70: The formats could locate in from common/utils/date-time

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+      })
+      .catch((e: Error) => {
+        setLoadingResourceError(e);
+        setLoading(false);
+      });
+  }, [resourceId]);
+
+  if (isLoading) {
+    return (
+      <>
```
minevala, line 99: Fragment is probably unnecessary here 🤔

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+          position="top-right"
+          autoClose
+          size="small"
+          label="Aukiolojakson lisääminen epäonnistui"
+          onClose={(): void => setSubmitStatus('init')}
+          type="error">
+          Aukiolojakson lisääminen epäonnistui. Yritä myöhemmin uudestaan.
+        </Notification>
+      )}
+      <div className="add-new-opening-period-page-header-row">
```
minevala, line 143: Naming suggestion: "add-new" -> "new"?
Or would just "opening-period-page-..." be enough?

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+            ref={register({ maxLength: 255 })}
+          />
+          <h3 className="opening-period-section-title">Ajanjakso</h3>
+          <section className="opening-period-time-period">
+            <Datepicker
+              id="openingPeriodBeginDate"
+              dataCy="openingPeriodBeginDate"
+              labelText="Alkaa"
+              onChange={(value): void => setPeriodBeginDate(value || null)}
+              onBlur={(): void => {
```
minevala, line 188: Are these empty blur-callbacks needed?

src/resource/resource-opening-hours/ResourceOpeningHours.tsx
```diff
@@ -63,8 +69,12 @@ const OpeningPeriodsList = ({
             value={language}
           />
           <Button
+            data-cy={addNewOpeningPeriodButtonDataCy}
```
minevala, line 72: I think we have used previously data-test-attribute 🤔

src/opening-period/page/CreateNewOpeningPeriodPage.scss
```diff
@@ -0,0 +1,122 @@
...
+  border: 2px solid rgb(128, 128, 128);
+  margin-bottom: $spacing-layout-2-xs;
+
+  @media screen and (max-width: $breakpoint-m) {
+    width: 100%;
+  }
+}
+
+.opening-period-section-title {
+  font-family: HelsinkiGrotesk-Bold, Arial, sans-serif;
```
minevala, line 46: Would it be easier to add into index.scss:
```css
h1, h2, h3, h4, h5, button, label, input {
   font-family: HelsinkiGrotesk-Bold, Arial, sans-serif;
}

```

minevala (commented):
This has required lots of coding 😅 
Good job 👍 
Few small comments.

I noticed also moderate axe-issue in the console:
> moderate: Page must contain a level-one heading https://dequeuniversity.com/rules/axe/3.5/page-has-heading-one?application=axeAPI

src/common/utils/api/api.ts
minevala: Or does it mean internal id?

src/common/utils/api/api.ts
VilleMiekkoja: - [ ] Move interface Resource under lib/types

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: If I remember correct, they were required by the Datepicker component. I'll check. Though now that we have copied the whole component, we could change them as optional 😁

src/resource/resource-opening-hours/ResourceOpeningHours.tsx
VilleMiekkoja: Good point! I'll change these data-cy ones to data-test to make them consistent with the ones we have used earlier. Also maybe data-test is better in that sense that it doesn't have to be an attribute just for e2e tests, but also for unit tests.

- [ ] Change data-cy -> data-test

src/opening-period/page/CreateNewOpeningPeriodPage.scss
VilleMiekkoja: Ah good point, specially if those all actually are with the same font-family! I'll check and will make the refactoring if that is the case. This was a good call!

- [ ] Check if common font-family is possible for h1, h2, h3, h4, h5, button, label, input elements

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: Good suggestion. I think in this case "opening-period-page-header-row" could be sufficient. Since these same styles will also probably be applied in the 'modify opening period page', it fits better for both purposes

src/common/utils/api/api.ts
VilleMiekkoja: I think it is the internal id here. Atleast the way it is applied in ResourcePage:

```
useEffect((): void => {
      if (resource) {
        api
          .getDatePeriod(resource.id)
          .then((ds: DatePeriod[]) => {
            setDatePeriods(ds);
          })
          .catch((e: Error) => {
            setError(e);
          });
      }
    }, [resource]);
```

And if we look at the Resource returned by api here: http://hauki-test.oc.hel.ninja/v1/resource/tprek:8215/ It shows:

image

It seems we can fetch a resource by applying the 'tprek:8215' format in the url, but then the resource that we get has an id with number format. I changed the format to number in the Resource type definition also.

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: - [ ] Remove unnecessary fragment

src/opening-period/page/CreateNewOpeningPeriodPage.scss
minevala: It might be actually:
```css
h1, h2, h3, h4 {
 font-family:  bold..
}

button, label {
 font-family: default..
}
```

VilleMiekkoja: Checked and it seems we could change the h-elements under common font-family. About the other elements, I fear there is a risk that it might affect also to all third party components (hds for example), and it could lead to some surprising side-effects?

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: - [ ] Remove unused onBlur methods

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: - [ ] Move these formats under common/utils/date-time

src/components/datepicker/Datepicker.tsx
VilleMiekkoja: - [ ] Move these formats under common/utils/date-time

VilleMiekkoja:
> This has required lots of coding 😅
> Good job 👍
> Few small comments.
> 
> I noticed also moderate axe-issue in the console:
> 
> > moderate: Page must contain a level-one heading https://dequeuniversity.com/rules/axe/3.5/page-has-heading-one?application=axeAPI

- [ ] Fix this accessibility problem by introducing a h1 element in the loader

minevala (approved):

PR #58, opened 2020-11-23
Hauki 222 loading indicator

minevala:
- Add LoadingIndicator-component according to HDS-guidelines: https://hds.hel.fi/components/spinner
- Use indicator in app-init and resource-page

Näyttökuva 2020-11-23 kello 11 22 30


Rikuoja (approved):
Looks good! But does starting the app (or loading the resource) really take more than three seconds? Perhaps on slow connections?

minevala:
Yes, most of the time, it is way faster, but sometimes individual requests can take more than three seconds.

cypress/integration/App.spec.ts
```diff
@@ -14,6 +14,8 @@ describe('Open aukiolot app', () => {
   it('Has no detectable a11y violations on resource page load', () => {
     cy.injectAxe();
     cy.get('[data-test=resource-info]', { timeout: 5000 }).should('be.visible');
+
+    cy.wait(1000); // Wait for HDS spinner cleanup.
```
VilleMiekkoja, line 18: I wonder if there's a better way to do this wait? Could we check that the spinner is no longer there? Or is there are problem in that because it might not have appeared yet, so we would get false positives? I wonder if there's a method in cypress for checking this with some initial waiting time.

src/index.scss
```diff
@@ -1,4 +1,5 @@
 @import '~hds-core/lib/base.css';
+@import '~hds-core/lib/components/loading-spinner/loading-spinner.css' ;
```
VilleMiekkoja, line 2: Are we required to separately import loading-spinner styles from hds? One could expect they come automatically with the component?

minevala: I'll remove this

cypress/integration/App.spec.ts
minevala: The clean-up callback is called one second after spinner disappears

VilleMiekkoja (approved):

VilleMiekkoja:
Good job! 💪

PR #59, opened 2020-11-24
Hauki 168 check permissions

minevala:
- Read tokens from query or storage in app init
- Replace PrivateRoute with PrivateResourceRoute
- PrivateResourceRoute checks the permissions before rendering the content


src/resource/PrivateResourceRoute.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { ReactNode, useEffect, useState } from 'react';
```
VilleMiekkoja, line 1: Would it be good to have some unit tests for this PrivateResourceRoute component?

src/common/utils/api/api.ts
```diff
@@ -140,4 +144,11 @@ export default {
     apiGet<AuthTestResponse>({
       path: authRequiredTest,
     }),
+
+  testPostPermission: async (resourceId: string): Promise<boolean> => {
```
VilleMiekkoja, line 148: Is this post permission specific to resource? If I understood correctly, it means the user has permission to make a post request to a specific resource? So not general postPermission, but resource-based? Though we currently only have posts to resource path?

VilleMiekkoja:
Looking good! I added couple comments

src/resource/PrivateResourceRoute.tsx
minevala: I will add 👍

src/common/utils/api/api.ts
minevala: Good point, I'll change the name

src/resource/PrivateResourceRoute.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface PrivateRouteProps extends RouteProps {
+  computedMatch?: ComputedMatch;
+  id: string;
+}
+
+export default function (routeProps: PrivateRouteProps): JSX.Element {
+  const { render, id, ...rest } = routeProps;
+  const resourceId = rest.computedMatch?.params.id;
+  const { authTokens, clearAuth }: Partial<AuthContextProps> = useAuth();
+  const [isLoading, setLoading] = useState<boolean>(true);
```
VilleMiekkoja, line 26: We discussed about these loaders on slack. I wonder if it is a good idea to introduce yet another loader to the application? Or have those other loaders been reduced at this pr also? :D. Just wondering because we've had quite many loaders and it makes the whole page load look very clumsy.

minevala: It actually replaces the one in the App.tsx

src/resource/PrivateResourceRoute.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    .spyOn(api, 'testResourcePostPermission')
+    .mockImplementation(() => Promise.resolve(hasPermission));
+};
+
+describe(`<PrivateResourceRoute />`, () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('should show loading indicator', async () => {
```
VilleMiekkoja, line 70: Should we use it or test? I think we have both in the codebase? 🤔

src/resource/PrivateResourceRoute.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    mockContext();
+    mockPermissionsApi(false);
+
+    const renderedComponent = renderRoutesWithPrivateRoute();
+
+    await act(async () => {
+      renderedComponent.update();
+    });
+
+    await act(async () => {
```
VilleMiekkoja, line 110: Just curious, why do we need double acts here?

minevala: Good question. First one is to mount the ResourceRoutes child component and the second one is to tick the useEffect-hook.

src/resource/PrivateResourceRoute.test.tsx
minevala: Which one do you prefer?

VilleMiekkoja: I've been using mostly 'it', but don't have a strong opinion on this :).

VilleMiekkoja (approved):

src/resource/PrivateResourceRoute.test.tsx
minevala: I can change these and other ones to it

PR #66, opened 2020-12-03
Hauki 117 edit page

minevala:
- Add edit-page for date-period
- Add edit-page-links to resource-page's date-period list items
- Add simple e2e-test for edit-page

Näyttökuva 2020-12-4 kello 10 47 30

Näyttökuva 2020-12-3 kello 16 15 33


cypress/support/commands.ts
```diff
@@ -58,3 +61,73 @@ Cypress.on('uncaught:exception', (err: Error): boolean | Error => {
...
+    resourceId: string;
+  }): Chainable => {
+    return cy
+      .visit('/')
+      .window()
+      .then((win: Window) => {
+        const apiUrl = win.ENV?.API_URL;
+
+        if (!apiUrl) {
+          return cy.task('log', 'API_URL is not set!!!');
```
VilleMiekkoja, line 87: I wonder if we could reduce the amount of nested promise chains by throwing an error here, and then enclosing this then branch?

src/common/utils/date-time/format.ts
```diff
@@ -19,3 +22,5 @@ export const formatDateRange = ({
 export const dateApiFormat = 'yyyy-MM-dd';
 export const dateFormFormat = 'dd.MM.yyyy';
 export const datetimeFormFormat = `${dateFormFormat} HH:mm`;
+export const transformToApiFormat = (formDate: string): string =>
```
VilleMiekkoja, line 25: Good idea to refactor this to here 👍

src/opening-period/page/EditOpeningPeriod.tsx
```diff
@@ -0,0 +1,301 @@
...
+  <Controller
+    control={control}
+    name={name}
+    render={({ value, onChange }): JSX.Element => (
+      <Datepicker
+        id={name}
+        dataTest={name}
+        labelText={label}
+        onChange={onChange}
+        value={new Date(value)}
```
VilleMiekkoja, line 47: Where does this value come from?

src/opening-period/page/EditOpeningPeriod.tsx
```diff
@@ -0,0 +1,301 @@
...
+          sv: null,
+          en: null,
+        },
+        start_date: transformToApiFormat(data.openingPeriodBeginDate),
+        end_date: transformToApiFormat(data.openingPeriodEndDate),
+        resource_state: ResourceState.OPEN,
+        override: false,
+        time_span_groups: [],
+      });
+      resetForm(updatedPeriod);
```
VilleMiekkoja, line 116: Is this resetForm necessary?

src/opening-period/page/EditOpeningPeriod.tsx
```diff
@@ -0,0 +1,301 @@
...
+      .then((r: Resource) => {
+        setResource(r);
+      })
+      .catch((e: Error) => {
+        setLoadingResourceError(e);
+        setIsLoading(false);
+      });
+  }, [resourceId]);
+
+  useEffect((): void => {
```
VilleMiekkoja, line 140: Is it best practice to have multiple useEffect calls, or to try to do everything inside one?

cypress/support/commands.ts
```diff
@@ -58,3 +61,73 @@ Cypress.on('uncaught:exception', (err: Error): boolean | Error => {
...
+
+        return cy
+          .exec('node ./scripts/generate-auth-params.js')
+          .then((params) =>
+            cy
+              .request(
+                'GET',
+                `${apiUrl}/v1/resource/${resourceId}?${params.stdout}`
+              )
+              .then((resourceResponse) => {
```
VilleMiekkoja, line 98: I think we could reduce the amount of nested promise chains here by closing the upper then before this line, and then we could continue with a new then that has the same intentation that the previous one has?

src/opening-period/page/EditOpeningPeriod.tsx
VilleMiekkoja: Maybe the updatedPeriod object could hold some api side validation errors potentially, and we would wish to render those and there resetForm would be required?

src/opening-period/page/EditOpeningPeriod.tsx
```diff
@@ -0,0 +1,271 @@
+import React, { useEffect, useState, useCallback } from 'react';
```
VilleMiekkoja, line 1: Comment about file name: Would it be better to include the word 'Page' in the names of these files EditOpeningPeriod.tsx and EditOpeningPeriod.scss. I know that the name 'page' is already present at the path of the component. But many times you only see the component name when you read the code. And it could be easier to read if you know from the name already what the component is about. I think we have it with ResourcePage and CreateNewOpeningPeriodPage as well 🤔

cypress/support/commands.ts
minevala: It throws an error now

src/opening-period/page/EditOpeningPeriod.tsx
minevala: it is the initial value of the react-hook-form which is undefined when defaultValues are not set

src/opening-period/page/EditOpeningPeriod.tsx
minevala: Oh, yes, I'll fix that

src/opening-period/page/EditOpeningPeriod.tsx
minevala: I think it is a good practice reset form state since API returns the updated value

src/opening-period/page/EditOpeningPeriod.tsx
minevala: This is fixed.

cypress/support/commands.ts
minevala: This is fixed.

src/opening-period/page/EditOpeningPeriod.tsx
minevala: Fixed

src/opening-period/page/EditOpeningPeriod.tsx
VilleMiekkoja: Though it should be the exact same value we put there? hmm

src/opening-period/page/EditOpeningPeriod.tsx
VilleMiekkoja: Good job with this!

VilleMiekkoja (approved):

VilleMiekkoja (approved):

PR #69, opened 2020-12-09
Hauki 118 add time span section to creation of date period form

VilleMiekkoja:
image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,7 @@ module.exports = {
     node: true,
   },
   rules: {
+    'react/no-array-index-key': 0,
     'react/prop-types': 0,
     'react/destructuring-assignment': 0,
     'react/static-property-placement': 0,
--- a/src/common/lib/types.ts
+++ b/src/common/lib/types.ts
@@ -34,7 +34,7 @@ export type TimeSpan = {
   description?: LanguageStrings;
   full_day?: boolean;
   resource_state?: ResourceState;
-  group?: 1;
+  group?: number;
 };
 
 export type TimeSpanFormFormat = {
--- a/src/common/lib/types.ts
+++ b/src/common/lib/types.ts
@@ -42,15 +42,7 @@ export type TimeSpanFormFormat = {
   endTime: string;
   startTime: string;
   resourceState: ResourceState;
-  weekdays: [
-    true | false,
-    true | false,
-    true | false,
-    true | false,
-    true | false,
-    true | false,
-    true | false
-  ];
+  weekdays: [boolean, boolean, boolean, boolean, boolean, boolean, boolean];
 };
 
 type ResourceStateApiOption = {
--- a/src/common/lib/types.ts
+++ b/src/common/lib/types.ts
@@ -22,26 +22,62 @@ enum WeekdayTypes {
 
 type Weekdays = Array<WeekdayTypes>;
 
-type TimeSpan = {
-  id: number;
-  created: string;
-  modified: string;
-  is_removed: boolean;
-  name: LanguageStrings;
-  description: LanguageStrings;
+export type TimeSpan = {
   start_time: string;
   end_time: string;
-  full_day: boolean;
   weekdays: Weekdays;
-  resource_state: ResourceState;
-  group: 1;
+  id?: number;
+  created?: string;
+  modified?: string;
+  is_removed?: boolean;
+  name?: LanguageStrings;
+  description?: LanguageStrings;
+  full_day?: boolean;
+  resource_state?: ResourceState;
+  group?: 1;
+};
+
+export type TimeSpanFormFormat = {
+  description: string;
+  endTime: string;
+  startTime: string;
+  resourceState: ResourceState;
+  weekdays: [
+    true | false,
+    true | false,
+    true | false,
+    true | false,
+    true | false,
+    true | false,
+    true | false
+  ];
+};
+
+type ResourceStateApiOption = {
+  value: string;
+  display_name: string;
+};
+
+export type ResourceStateOption = {
+  label: string;
+  value: string;
+};
+
+export type DatePeriodOptions = {
+  actions: {
+    POST: {
+      resource_state: {
+        choices: ResourceStateApiOption[];
+      };
+    };
+  };
 };
 
 export type TimeSpanGroup = {
-  id: number;
+  id?: number;
   time_spans: TimeSpan[];
   rules: [];
-  period: number;
+  period?: number;
 };
 
 export type DatePeriod = {
@@ -53,7 +89,7 @@ export type DatePeriod = {
   description: LanguageStrings;
   start_date: string;
   end_date: string;
-  resource_state: ResourceState;
+  resource_state?: ResourceState;
   override: boolean;
   resource: number;
   time_span_groups: TimeSpanGroup[];
--- a/src/common/lib/types.ts
+++ b/src/common/lib/types.ts
@@ -45,7 +45,7 @@ export type TimeSpanFormFormat = {
   weekdays: [boolean, boolean, boolean, boolean, boolean, boolean, boolean];
 };
 
-type ResourceStateApiOption = {
+export type ResourceStateApiOption = {
   value: string;
   display_name: string;
 };
--- a/src/common/utils/api/api.ts
+++ b/src/common/utils/api/api.ts
@@ -47,7 +47,6 @@ interface DataRequestParameters {
 
 interface OptionsParameters {
   path: string;
-  useRootPath?: boolean;
 }
 
 interface PostParameters extends DataRequestParameters {
@@ -149,12 +148,9 @@ async function apiPut<T>({
   });
 }
 
-async function apiOptions<T>({
-  path,
-  useRootPath = false,
-}: OptionsParameters): Promise<T> {
+async function apiOptions<T>({ path }: OptionsParameters): Promise<T> {
   return request<T>({
-    url: `${apiBaseUrl}${useRootPath ? '' : '/v1'}${path}/`,
+    url: `${apiBaseUrl}/v1${path}/`,
     headers: {
       'Content-Type': 'application/json',
     },
--- a/src/common/utils/api/api.ts
+++ b/src/common/utils/api/api.ts
@@ -3,6 +3,7 @@ import * as querystring from 'querystring';
 import { ParsedUrlQueryInput } from 'querystring';
 import {
   DatePeriod,
+  DatePeriodOptions,
   LanguageStrings,
   Resource,
   ResourceState,
@@ -42,6 +43,11 @@ interface DataRequestParameters {
   path: string;
   headers?: { [key: string]: string };
   data?: RequestParameters;
+}
+
+interface OptionsParameters {
+  path: string;
+  useRootPath?: boolean;
 }
 
 interface PostParameters extends DataRequestParameters {
@@ -143,6 +149,22 @@ async function apiPut<T>({
   });
 }
 
+async function apiOptions<T>({
+  path,
+  useRootPath = false,
+}: OptionsParameters): Promise<T> {
+  return request<T>({
+    url: `${apiBaseUrl}${useRootPath ? '' : '/v1'}${path}/`,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    method: 'options',
+    validateStatus(status) {
+      return status < 300;
+    },
+  });
+}
+
 interface AuthTestResponse {
   message: string;
   username: string;
@@ -178,6 +200,11 @@ export default {
   getDatePeriod: (datePeriodId: number): Promise<DatePeriod> =>
     apiGet<DatePeriod>({
       path: `${datePeriodBasePath}/${datePeriodId}`,
+    }),
+
+  getDatePeriodFormOptions: (): Promise<DatePeriodOptions> =>
+    apiOptions<DatePeriodOptions>({
+      path: `${datePeriodBasePath}`,
     }),
 
   postDatePeriod: (datePeriod: DatePeriod): Promise<DatePeriod> =>
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -1,0 +1,37 @@
+import { Button as HDSButton } from 'hds-react';
+import React, { ReactNode } from 'react';
+import './Button.scss';
+
+type ButtonVariant = 'primary' | 'secondary';
+type ButtonTypeVariant = 'button' | 'submit' | 'reset' | undefined;
+
+interface ButtonProps {
+  children: ReactNode;
+  onClick?: () => void;
+  variant?: ButtonVariant;
+  dataTest?: string;
+  className?: string;
+  type?: ButtonTypeVariant;
+}
+
+export default function Button({
+  variant = 'primary',
+  children,
+  dataTest,
+  onClick,
+  className = '',
+  type = 'button',
+}: ButtonProps): JSX.Element {
+  return (
+    <HDSButton
+      data-test={dataTest}
+      className={`button-common ${
+        variant === 'primary' ? 'primary-button' : 'secondary-button'
+      } ${className}`}
+      variant={variant}
+      onClick={onClick}
+      type={type}>
+      {children}
+    </HDSButton>
+  );
+}
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -14,8 +14,7 @@ interface ButtonProps {
   type?: ButtonTypeVariant;
 }
 
-export default function Button({
-  variant = 'primary',
+export function SecondaryButton({
   children,
   dataTest,
   onClick,
@@ -25,10 +24,27 @@ export default function Button({
   return (
     <HDSButton
       data-test={dataTest}
-      className={`button-common ${
-        variant === 'primary' ? 'primary-button' : 'secondary-button'
-      } ${className}`}
-      variant={variant}
+      className={`button-common secondary-button ${className}`}
+      variant="secondary"
+      onClick={onClick}
+      type={type}>
+      {children}
+    </HDSButton>
+  );
+}
+
+export function PrimaryButton({
+  children,
+  dataTest,
+  onClick,
+  className = '',
+  type = 'button',
+}: ButtonProps): JSX.Element {
+  return (
+    <HDSButton
+      data-test={dataTest}
+      className={`button-common primary-button ${className}`}
+      variant="primary"
       onClick={onClick}
       type={type}>
       {children}
--- a/src/index.scss
+++ b/src/index.scss
@@ -30,14 +30,10 @@ h5 {
 
 fieldset {
   padding: 0 !important;
-  margin-inline-start: 0;
-  margin-inline-end: 0;
-  padding-block-start: 0;
-  padding-block-end: 0;
-  padding-inline-start: 0;
-  padding-inline-end: 0;
+  margin-right: 0;
+  margin-left: 0;
   border: none;
-  min-inline-size: min-content;
+  min-width: min-content;
 }
 
 #root {
--- a/src/index.scss
+++ b/src/index.scss
@@ -36,6 +36,10 @@ fieldset {
   min-width: min-content;
 }
 
+label {
+  font-family: HelsinkiGrotesk-Medium, Arial, sans-serif;
+}
+
 #root {
   margin: 0;
   height: 100%;
--- a/src/index.scss
+++ b/src/index.scss
@@ -28,6 +28,18 @@ h5 {
   font-family: HelsinkiGrotesk-Bold, Arial, sans-serif;
 }
 
+fieldset {
+  padding: 0 !important;
+  margin-inline-start: 0;
+  margin-inline-end: 0;
+  padding-block-start: 0;
+  padding-block-end: 0;
+  padding-inline-start: 0;
+  padding-inline-end: 0;
+  border: none;
+  min-inline-size: min-content;
+}
+
 #root {
   margin: 0;
   height: 100%;
--- a/src/opening-period/description/OpeningPeriodDescription.scss
+++ b/src/opening-period/description/OpeningPeriodDescription.scss
@@ -13,7 +13,6 @@
 .opening-period-optional-description {
   padding: $spacing-layout-2-xs;
   width: calc(0.5808333 * 100%);
-  height: 164px;
   border: 2px solid rgb(128, 128, 128);
   margin-bottom: $spacing-layout-2-xs;
 
--- a/src/opening-period/description/OpeningPeriodDescription.scss
+++ b/src/opening-period/description/OpeningPeriodDescription.scss
@@ -1,20 +1,12 @@
 @import 'node_modules/hds-design-tokens/lib/all.scss';
 
 .add-new-opening-period-title {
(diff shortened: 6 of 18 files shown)


.eslintrc.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
     node: true,
   },
   rules: {
+    'react/no-array-index-key': 0,
```
VilleMiekkoja, line 22: Added this because of src/opening-period/page/CreateNewOpeningPeriodPage.tsx line 252:

```
            <h3 className="opening-period-section-title">Aukioloajat</h3>
            {fields.map((item, index) => (
              <TimeSpan
                resourceStateOptions={resourceStateOptions}
                setValue={setValue}
                register={register}
                key={`time-span-${index}`}
                index={index}
                remove={remove}
              />
```

Without the rule lint will warn that applying index as part of key is not good. We should apply the unique identifier of the thing that we are looping. The problem is, the time span has not yet being created, so it does not yet have a unique identifier that we could apply. So I saw no other way than to apply the index

src/components/button/Button.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children: ReactNode;
+  onClick?: () => void;
+  variant?: ButtonVariant;
+  dataTest?: string;
+  className?: string;
+  type?: ButtonTypeVariant;
+}
+
+export default function Button({
+  variant = 'primary',
```
VilleMiekkoja, line 18: Should we make separate components for _PrimaryButton_ and _SecondaryButton_? Or do you think applying the variant prop is enough? I think we might need to make a third kind of button also, so this might not scale very well in the long run. But maybe we think about that when the time comes?

src/index.scss
```diff
@@ -28,6 +28,18 @@ h5 {
   font-family: HelsinkiGrotesk-Bold, Arial, sans-serif;
 }
 
+fieldset {
```
VilleMiekkoja, line 31: These are set to try to reset the default css values for fieldset that it gets from browsers.

src/common/lib/types.ts
```diff
@@ -22,26 +22,62 @@ enum WeekdayTypes {
...
+      resource_state: {
+        choices: ResourceStateApiOption[];
+      };
+    };
+  };
 };
 
 export type TimeSpanGroup = {
-  id: number;
+  id?: number;
```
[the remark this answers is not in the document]
minevala: I think we need to add NewDatePeriod-type with optional id. Then we could have the normal DatePeriod-type with mandatory id. I'm not sure if we need to do the same for TimeSpanGroup tough 🤔

src/components/button/Button.tsx
minevala: It might be better to split those two. I was already thinking icon-button etc. 😁

src/index.scss
minevala: I'm not sure about the browser support of CSS logical properties 🤔

src/opening-period/page/CreateNewOpeningPeriodPage.scss
```diff
@@ -16,45 +16,26 @@
   label {
     display: block;
     margin-top: $spacing-m;
+    font-family: HelsinkiGrotesk-Medium, Arial, sans-serif;
```
minevala, line 19: Would it make sense to put this into index.scss as a global style

src/opening-period/page/CreateNewOpeningPeriodPage.scss
```diff
@@ -16,45 +16,26 @@
...
-  border: 2px solid rgb(128, 128, 128);
-  margin-bottom: $spacing-layout-2-xs;
-
-  @media screen and (max-width: $breakpoint-m) {
-    width: 100%;
-  }
-}
-
 .opening-period-section-title {
+  font-family: HelsinkiGrotesk-Bold, Arial, sans-serif;
```
minevala, line 35: Is this needed?  There seems to be the same CSS rule in index.scss for h3-element

src/opening-period/page/CreateNewOpeningPeriodPage.scss
```diff
@@ -67,48 +48,36 @@
...
+  margin-top: $spacing-l;
+}
 
-  &:focus {
-    border-color: var(--color-coat-of-arms) !important;
-    outline-color: var(--color-coat-of-arms);
-    color: var(--color-coat-of-arms) !important;
-  }
+.time-span-group {
+  border-top: $spacing-3-xs solid var(--color-coat-of-arms);
```
minevala, line 82: Maybe this color could be in variables.css?

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -9,37 +9,84 @@ import { ResourceInfo } from '../../resource/page/ResourcePage';
...
+    return {
+      description: {
+        fi: timeSpan.description,
+        sv: null,
+        en: null,
+      },
+      end_time: `${timeSpan.endTime}:00`,
+      start_time: `${timeSpan.startTime}:00`,
+      resource_state: timeSpan.resourceState,
+      weekdays: timeSpan.weekdays.reduce(
```
minevala, line 50: JS native map has index as a second argument in the callback function. Would be easier to use that in the render function? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -73,30 +120,46 @@ export default function CreateNewOpeningPeriodPage({
...
+              label: optionInApiFormat.display_name,
+            };
+          })
+        );
         setLoading(false);
       })
       .catch((e: Error) => {
         setLoadingResourceError(e);
         setLoading(false);
       });
```
minevala, line 161: This can be written also in this kind of format:
```js
const fetchData = async (): Promise<void> => {
      try {
        const [apiResource, datePeriod] = await Promise.all([
          api.getResource(resourceId),
          api.getDatePeriod(id),
        ]);
        setResource(apiResource);
        resetForm(datePeriod);
        setIsLoading(false);
      } catch (e) {
        setHasDataLoadingError(e);
        setIsLoading(false);
      }
    };

    fetchData();
```

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -73,30 +120,46 @@ export default function CreateNewOpeningPeriodPage({
...
-  }, [resourceId]);
+  }, [append, resourceId]);
```
minevala, line 162: I'm not sure if this append is needed 🤔

src/opening-period/time-span/TimeSpan.scss
```diff
@@ -0,0 +1,45 @@
...
+  color: var(--color-coat-of-arms);
+}
+
+.start-and-end-time-container {
+  display: flex;
+  align-items: flex-end;
+}
+
+.dash-between-start-and-end-time-container {
+  height: 56px;
```
minevala, line 13: This seem to be redundant declaration

src/opening-period/time-span/TimeSpan.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+.time-span-first-header-row {
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-end;
+}
+
+.time-span-container {
+  border-top: $spacing-4-xs solid var(--color-coat-of-arms);
```
minevala, line 43: Maybe this color could be in the variables.css?

src/opening-period/time-span/TimeSpan.scss
```diff
@@ -0,0 +1,45 @@
...
+  margin-top: 46px;
```
minevala, line 44: Could this be replaced with some predefined HDS-spacing?

src/opening-period/time-span/TimeSpan.tsx
```diff
@@ -0,0 +1,94 @@
...
+  remove,
+  register,
+  setValue,
+  resourceStateOptions,
+}: {
+  index: number;
+  remove: Function;
+  register: Function;
+  setValue: Function;
+  resourceStateOptions: ResourceStateOption[] | undefined;
```
minevala, line 19: Should it be mandatory to have resourceStateOptions in this component? Otherwise we would not render it at all

src/opening-period/time-span/TimeSpan.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!resourceStateOptions) {
+    throw new Error('This should never happen but typescript');
+  }
+
+  return (
+    <div data-test={`time-span-${index}`} className="time-span-container">
+      <div className="time-span-first-header-row">
+        <Weekdays index={index} register={register} />
+        <HDSButton
```
minevala, line 33: Can this be replaced with our button implementation?

src/index.scss
VilleMiekkoja: Good catch, maybe we could just replace those with the regular margin-right and margin-left?

src/opening-period/time-span/Weekdays.tsx
```diff
@@ -0,0 +1,107 @@
...
+        htmlFor={`timeSpans[${index}].weekdays[0]`}
+        className="day-label">
+        <input
+          ref={register()}
+          type="checkbox"
+          name={`timeSpans[${index}].weekdays[0]`}
+          className="opening-hours-day-checkbox-input"
+          id={`timeSpans[${index}].weekdays[0]`}
+        />
+        <span className="day-option">Ma</span>
```
minevala, line 25: It might be easier to maintain this if there would be a DayCheckBox-component, etc.
Then it would be just:
```js
<fieldset>
      <legend>Päivät *</legend>
      <DayCheckbox day="monday"/>
      <DayCheckbox day="tuesday"/>
      ....
</fieldset>
```

minevala: It did not focus on these day-checkboxes when I tested navigating the form with just tab 🤔

minevala (commented):
I tested the thingy and it works 👍 
Nice work. Some small comments here and there.

.eslintrc.js
minevala: ✅

src/common/lib/types.ts
```diff
@@ -22,26 +22,62 @@ enum WeekdayTypes {
...
-  group: 1;
+  id?: number;
+  created?: string;
+  modified?: string;
+  is_removed?: boolean;
+  name?: LanguageStrings;
+  description?: LanguageStrings;
+  full_day?: boolean;
+  resource_state?: ResourceState;
+  group?: 1;
```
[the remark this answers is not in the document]
VilleMiekkoja: Probably yes. Think I just blindly cargo-culted it, good catch!

src/common/lib/types.ts
```diff
@@ -22,26 +22,62 @@ enum WeekdayTypes {
...
+};
+
+export type TimeSpanFormFormat = {
+  description: string;
+  endTime: string;
+  startTime: string;
+  resourceState: ResourceState;
+  weekdays: [
+    true | false,
```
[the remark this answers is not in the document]
VilleMiekkoja: Haha, good catch! I blame late night typescript wars :see_no_evil:

src/common/lib/types.ts
```diff
@@ -22,26 +22,62 @@ enum WeekdayTypes {
...
+};
+
+export type ResourceStateOption = {
+  label: string;
+  value: string;
+};
+
+export type DatePeriodOptions = {
+  actions: {
+    POST: {
```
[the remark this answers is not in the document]
VilleMiekkoja: Yes correct, it is the format of the response. I agree it's a bit complicated. Hmm. should we rename it better?

src/common/utils/api/api.ts
```diff
@@ -123,6 +129,22 @@ async function apiPost<T>({
   });
 }
 
+async function apiOptions<T>({
+  path,
+  useRootPath = false,
+}: OptionsParameters): Promise<T> {
+  return request<T>({
+    url: `${apiBaseUrl}${useRootPath ? '' : '/v1'}${path}/`,
```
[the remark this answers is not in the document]
VilleMiekkoja: Good catch! I might have cargo-culted it from there 😅

src/opening-period/description/OpeningPeriodDescription.scss
```diff
@@ -0,0 +1,23 @@
+@import 'node_modules/hds-design-tokens/lib/all.scss';
+
+.add-new-opening-period-title {
+  font-family: HelsinkiGrotesk-Regular, Arial, sans-serif;
+  font-size: $fontsize-body-l;
```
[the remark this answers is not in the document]
VilleMiekkoja: Good find, I'll check that out!

src/common/lib/types.ts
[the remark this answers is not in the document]
minevala: I guess it is fine like this but now I'm thinking if there are different options for PUT/PATCH. Then it feels a bit too complicated.

src/opening-period/time-span/Weekdays.tsx
VilleMiekkoja: Clearly a bug, I'll try to fix that, good find!

- [ ] Enable focus to the day-checkboxes when navigating by tab

VilleMiekkoja: > It might be easier to maintain this if there would be a DayCheckBox-component, etc.
> Then it would be just:
> 
> ```js
> <fieldset>
>       <legend>Päivät *</legend>
>       <DayCheckbox day="monday"/>
>       <DayCheckbox day="tuesday"/>
>       ....
> </fieldset>
> ```

Good idea, I'll extract that as a separate component!

- [ ] Extract DayCheckBox as a separate component

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: Js native map will probably create as many elements as there are in the original array? In the api format the array length is reduced; it consists of only those days that are selected

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: True, I considered it but it creates a new unnecessary function and I just thought plain promise version was cleaner 🤔

src/opening-period/time-span/TimeSpan.scss
VilleMiekkoja: There's $spacing-2-xl which is 48px. It's close, I'll check if we could apply that!

- [ ] Check if we can apply $spacing-2-xl instead of hard-coding

src/opening-period/description/OpeningPeriodDescription.scss
```diff
@@ -0,0 +1,23 @@
...
+  padding-left: $spacing-s;
+  padding-right: $spacing-s;
+  border: 2px solid rgb(128, 128, 128);
+  min-width: 400px;
+}
+
+.opening-period-optional-description {
+  padding: $spacing-layout-2-xs;
+  width: calc(0.5808333 * 100%);
+  height: 164px;
```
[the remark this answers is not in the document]
VilleMiekkoja: Good call, I'll check if we could replace the height completely with the use of rows-attribute!

- [ ] Try to apply rows-attribute instead of hard-coding height

src/opening-period/page/CreateNewOpeningPeriodPage.scss
VilleMiekkoja: I can try that. It might be challenging because often HDS styles overrides our styles that are not specific enough. But I'll see what we can do. 

- [ ] Check if this could be put into index.scss as a global style

src/opening-period/page/CreateNewOpeningPeriodPage.scss
VilleMiekkoja: - [ ] Check if this font-family style definition is needed, and remove it if it is not

src/components/button/Button.tsx
VilleMiekkoja: - [ ] Create separate components for _PrimaryButton_ and _SecondaryButton_

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
VilleMiekkoja: Good catch, no idea why that got left there!

- [ ] Remove unneeded dependency append from useEffect dependency array

src/opening-period/time-span/TimeSpan.tsx
VilleMiekkoja: At the moment, no. Because our button only supports primary and secondary types. We could make a separate supplementary button, but because this was the first time when it was needed, I did not yet refactor it to a separate component, to avoid abstraction. But once we have the need for it, I think it should be extracted

src/opening-period/time-span/TimeSpan.scss
VilleMiekkoja: - [ ] Remove redundant height style declaration

src/opening-period/description/OpeningPeriodDescription.scss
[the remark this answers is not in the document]
VilleMiekkoja: Success, was able to successfully replace the height with rows attribute!

src/common/lib/types.ts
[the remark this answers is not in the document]
VilleMiekkoja: We probably need to think about this when we combine the edit date period page with this new form. I suggest we deal with the issue during that. We might be able to apply the same structure there too? What do you say?

src/opening-period/page/CreateNewOpeningPeriodPage.scss
minevala: I thought that HDS did not style labels, and that is why we needed to have our styles for them 🤔

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
minevala: This is resolved then

src/common/lib/types.ts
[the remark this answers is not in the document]
minevala: Ok, I just hope we can easily combine this to edit without excess overhead

src/opening-period/time-span/Weekdays.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <label
+      data-test={dataTest}
+      htmlFor={`timeSpans[${timeSpanIndex}].weekdays[${dayIndex}]`}
+      className="day-label">
+      <input
+        ref={register()}
+        type="checkbox"
+        name={`timeSpans[${timeSpanIndex}].weekdays[${dayIndex}]`}
+        className="opening-hours-day-checkbox-input"
```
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Remove redundant className 'opening-hours-day-checkbox-input'

src/opening-period/time-span/TimeSpan.scss
VilleMiekkoja: The most fitting by name seemed to be this: `--primary-color: var(--color-coat-of-arms);` Maybe we use that?

- [ ] Apply the 'primary-color' variable found in variables.css

VilleMiekkoja: Damn I think I understood the question wrong. You were suggesting that we would put that color there, not that it might already be there?

src/opening-period/description/OpeningPeriodDescription.scss
```diff
@@ -0,0 +1,23 @@
+@import 'node_modules/hds-design-tokens/lib/all.scss';
+
+.add-new-opening-period-title {
+  font-family: HelsinkiGrotesk-Regular, Arial, sans-serif;
+  font-size: $fontsize-body-l;
+  line-height: 2.5;
```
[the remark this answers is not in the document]
VilleMiekkoja: Those line-heights seem to be too small:

image

src/opening-period/time-span/TimeSpan.scss
minevala: Yes, but I guess it is ok to use this as well

minevala: I think primary color says it

src/opening-period/description/OpeningPeriodDescription.scss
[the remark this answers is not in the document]
minevala: Or should this be replaced with HDS-text-input?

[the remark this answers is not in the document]
minevala: Does it play well with react-form-hook?

src/opening-period/page/CreateNewOpeningPeriodPage.scss
VilleMiekkoja: It seemed to be redundant, so I removed it!

src/opening-period/description/OpeningPeriodDescription.scss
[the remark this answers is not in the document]
VilleMiekkoja: We probably could replace it with HDS-text-input element, though it's little outside of the scope of this PR, since the purpose was to continue the form, even though I did some refactoring to the form. Maybe I'll create a jira issue about it?

src/opening-period/time-span/TimeSpan.scss
VilleMiekkoja: $spacing-2-xl was close enough, and I checked that the UI also seemed to look ok with it, so applied that!

src/opening-period/description/OpeningPeriodDescription.scss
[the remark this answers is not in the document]
VilleMiekkoja: I checked the HDS base css helper class, and it seems the font-size helper class also comes with line-height definition. And unfortunately it's incorrect line-height for us. Atleast as long as we have our own input component

src/opening-period/page/CreateNewOpeningPeriodPage.scss
VilleMiekkoja: I put it to index.scss. Looked UI locally and it felt normal, so let's go with that!

src/opening-period/description/OpeningPeriodDescription.scss
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.opening-period-optional-description {
+  padding: $spacing-layout-2-xs;
+  width: calc(0.5808333 * 100%);
+  height: 164px;
+  border: 2px solid rgb(128, 128, 128);
+  margin-bottom: $spacing-layout-2-xs;
+
+  @media screen and (max-width: $breakpoint-m) {
```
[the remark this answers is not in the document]
VilleMiekkoja: I looked into it and didn't figure out a way to do it this time. Also these styles might change soon. We discussed to leave it as it is for now.

src/opening-period/time-span/TimeSpan.tsx
VilleMiekkoja: If I remove that `| undefined`, then typescript will complain in CreateNewOpeningPeriodPage.tsx this:

image

minevala (approved):
Great job 🎉

PR #71, opened 2020-12-11
Hauki 260 delete date period

minevala:
- Add custom confirmation modal (HDS will provide modal in the future)
- Add delete-buttons into date-periods in resource-page (button click will open a confirmation modal, confirm will delete date-period)

Näyttökuva 2020-12-11 kello 11 43 01
Näyttökuva 2020-12-11 kello 11 34 59


cypress/integration/DeleteOpeningPeriod.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    cy.get('[data-test=resource-opening-periods-list]', {
+      timeout: 5000,
+    })
+      .find(`[data-test="openingPeriodDeleteLink-${dataPeriodId}"]`)
+      .click({ log: true });
+
+    // Check that delete modal exists
+    cy.get('[data-test=modalTitle]', {
+      timeout: 5000,
+    }).should('have.text', 'Oletko varma että haluat poistaa aukiolojakson?');
```
VilleMiekkoja, line 41: I feel like this might be unnecessarily strict check? If we wish to change the text, then this test will fail, even our application will still work ok. I think it would be better to just verify the existence of the element with that data-test id, and leave it at that probably. If there's a more specific element we would want to test, then maybe we could give the data-test attribute to that element?

minevala: I could just see if the title exists 🤔

src/components/modal/Modal.scss
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.modal-content {
+  line-height: var(--lineheight-l);
+  margin-top: var(--spacing-s);
+}
+
+.modal-actions {
+  margin-top: var(--spacing-xl);
+  display: grid;
```
VilleMiekkoja, line 60: Grid for displaying two buttons side by side seems little overkill? It feels like there's an easier way to do this, without introducing the complex concept of grid here? 🤔

minevala: Maybe 🤔 😁  Basic grid is not complex at all but I'll check if there is other way to do this.
Notice that the buttons width was in design 50% - margin in between (no matter how long the content id)

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,121 @@
+import React, { ReactNode, RefObject, useEffect, useState } from 'react';
+import ReactDOM from 'react-dom';
+import { IconAlertCircle, IconCross } from 'hds-react';
+import { PrimaryButton, SecondaryButton } from '../button/Button';
+import './Modal.scss';
+
+const ModalBackDrop = (): JSX.Element => <div className="modal-backdrop" />;
```
VilleMiekkoja, line 7: Im not sure if extracting this ModalBackDrop here is a good idea? It increases the jumps reader has to do while reading the code, yet it's so small in size that I feel we could maybe just inline it directly to the code?

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,121 @@
...
+export const useModal = (): UseModalProps => {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const toggle = (): void => setIsOpen(!isOpen);
+  return {
+    isOpen,
+    toggle,
+  };
+};
+
+export function ConfirmationModal({
```
VilleMiekkoja, line 41: I wonder if it would be a good idea to match the name of the exported function with the filename of the component? 🤔

minevala: I can rename the file

src/components/modal/Modal.tsx
minevala: I will inline this.

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,121 @@
...
+type UseModalProps = {
+  isOpen: boolean;
+  toggle: () => void;
+};
+
+export const useModal = (): UseModalProps => {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const toggle = (): void => setIsOpen(!isOpen);
+  return {
+    isOpen,
```
VilleMiekkoja, line 36: Since these variables and functions are applied outside of this component, I wonder would it be better to name them more descriptive? isModalOpen and toggleModal? 🤔

minevala: Will rename.

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const onKeyDown = (event: KeyboardEvent): void => {
+    if (event.key === 'Escape') {
+      onClose();
+    }
+  };
+
+  useEffect(() => {
+    if (isOpen) {
+      document.body.classList.add('modal-visible');
```
VilleMiekkoja: This feels like not a react way of doing dom manipulation but I guess it is the only good way to do it in this situation 🤔

minevala: There might be some other way but this was the easiest 😁

VilleMiekkoja:
Not sure if it's by design, but it feels like the modal is not fully centered horizontally? While I feel like it applies to desktop screen size too, the offset can be seen more easily on smaller screensize:

image

src/index.scss
```diff
@@ -54,3 +54,21 @@ code {
...
+.button-icon {
+  background: none;
+  border: none;
+  color: inherit;
+  cursor: pointer;
+  font: inherit;
+  padding: var(--spacing-xs) var(--spacing-s);
+
+  &:hover {
+    background-color: var(--opening-period-listing-button-hover-background-color);
```
VilleMiekkoja, line 67: Maybe we need to rename this color as it's now grown as a more generic color applied in many components?

minevala: Added a specific variable

src/resource/resource-opening-hours/ResourceOpeningHours.tsx
```diff
@@ -112,10 +107,55 @@ export default function ResourceOpeningHours({
...
+export default function ResourceOpeningHours({
+  resourceId,
+}: {
+  resourceId: number;
+}): JSX.Element | null {
+  const [error, setError] = useState<Error | undefined>(undefined);
+  const [[defaultPeriods, exceptionPeriods], setDividedDatePeriods] = useState<
+    DatePeriod[][]
+  >([[], []]);
+  const fetchDatePeriods = async (id: number): Promise<void> => {
```
VilleMiekkoja, line 119: When re-render of this component occurs, are all of these functions created again? If that's the case Im wondering if it would be wiser to define them outside of this component?

minevala: Yes, it might be. It probably has to take a callback as parameter 🤔

minevala: Probably but I'm not sure does it cause perf issues?

VilleMiekkoja (approved):
Well done! 🥇

PR #78, opened 2020-12-16
Hauki 226 merge open period form

minevala:
- Add shared opening-period form (for create and edit date-period pages)
- Improve e2e-tests
- Add ResourceInfoSubtitle-component (to get rid of duplicate styles)
- Remove other duplicate styles

localhost_3000_resource_34299_period_new (1)


Rikuoja (approved):
The picture looks good!

src/common/lib/types.ts
```diff
@@ -46,12 +46,23 @@ export type TimeSpan = {
...
+  boolean,
+  boolean,
+  boolean,
+  boolean,
+  boolean,
+  boolean
+];
+
 export type TimeSpanFormFormat = {
+  id?: number | string;
```
VilleMiekkoja, line 60: Can it be both, string or number?

minevala: Probably no, I'll fix

src/common/utils/date-time/format.ts
```diff
@@ -8,15 +8,17 @@ export const formatDateRange = ({
...
-  startDate: string;
-  endDate: string;
+  startDate?: string;
+  endDate?: string;
 }): string => {
-  const isSingleDate: boolean = !endDate || startDate === endDate;
-  const formattedStartDate: string = formatDate(startDate);
-  if (isSingleDate) {
-    return formattedStartDate;
+  if (!startDate || (!startDate && !endDate)) {
```
VilleMiekkoja, line 14: I noticed we don't yet have unit test for this function. It looks the kind of function that might be easy to unit test, so I wonder if it would be worth it?

minevala: I'll add a test

src/opening-period/description/OpeningPeriodDescription.tsx
```diff
@@ -21,31 +20,37 @@ export default function OpeningPeriodDescription({
...
-      </label>
-      <TextArea
-        cols={90}
-        rows={9}
-        className="opening-period-optional-description"
-        id="openingPeriodOptionalDescription"
-        name="openingPeriodOptionalDescription"
-        ref={register({ maxLength: 255 })}
-      />
+      <div className="form-control">
```
VilleMiekkoja, line 23: I think introducing this form-control helper div and class is actually a good idea. The margin-up was previously handled by styling on the labels, but this feels better and more right way to do it 👍

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,322 @@
...
+} from '../../common/lib/types';
+import { transformToApiFormat } from '../../common/utils/date-time/format';
+import Datepicker from '../../components/datepicker/Datepicker';
+import { PrimaryButton, SecondaryButton } from '../../components/button/Button';
+import { ErrorToast, SuccessToast } from '../../components/notification/Toast';
+import OpeningPeriodDescription from '../description/OpeningPeriodDescription';
+import TimeSpan from '../time-span/TimeSpan';
+import './OpeningPeriodForm.scss';
+
+type OpeningPeriodDataBase = {
```
VilleMiekkoja, line 23: I wonder if OpeningPeriodBaseData would be better? DataBase feels like it's little too close to a word database, which has little different meaning I think 😁

minevala: Yes, good idea 😁

src/opening-period/description/OpeningPeriodDescription.tsx
minevala: I was thinking of separate component but then I was not sure where to but it.

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,322 @@
...
+  openingPeriodTitle: string;
+  openingPeriodOptionalDescription: string;
+  openingPeriodBeginDate: string | undefined;
+  openingPeriodEndDate: string | undefined;
+};
+
+interface OpeningPeriodFormData extends OpeningPeriodDataBase {
+  timeSpans: Array<TimeSpanFormFormat> | {}[];
```
VilleMiekkoja, line 31: Just for interest and to improve my typescript skills, can you tell me what that `{}[]` means? :)

minevala: It accepts an empty object array as well. We need that in as a timeSpan default value. I was struggling with this as well 😅

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,322 @@
...
+}
+
+export interface OpeningPeriodApiData extends OpeningPeriodDataBase {
+  timeSpans: Array<TimeSpanApiFormat>;
```
VilleMiekkoja, line 35: I think in api data the names are in snake_case? So I wonder why timeSpans is in camelCase here?

src/opening-period/description/OpeningPeriodDescription.tsx
minevala: Maybe component/forms/Control.tsx ?

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Good question, I check this one 🤔

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,322 @@
...
+  submitFn,
+  successTexts,
+  errorTexts,
+}: {
+  formId: string;
+  datePeriod?: DatePeriod;
+  resourceId: number;
+  datePeriodOptions: DatePeriodOptions;
+  submitFn: (datePeriod: DatePeriod) => Promise<DatePeriod>;
+  successTexts: NotificationTexts;
```
VilleMiekkoja, line 138: I wonder would it be more explicit to have separate props for successLabel and successText? Something like submitSuccessNotificationText and submitSuccessNotificationLabel hmm.. Maybe same for errorTexts?

VilleMiekkoja: Then again Im not sure if it's better to increase the amount of props. One option would be to just rename the successTexts to successTextAndLabel.

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,322 @@
...
+      ),
+    };
+  });
+}
+
+const dropMilliseconds = (time: string): string => time.slice(0, -3);
+const getLowestWeekdayNumber = (timeSpan: TimeSpanApiFormat): number =>
+  timeSpan && timeSpan.weekdays ? timeSpan.weekdays.sort()[0] : 0;
+
+function formatApiTimeSpansToFormFormat(
```
VilleMiekkoja, line 76: I wonder if it would be a good idea to add unit tests for these formatFunctions `formatApiTimeSpansToFormFormat`and `formatTimeSpansToApiFormat`. They seem like functions that would be easy and handy to unit test.

src/opening-period/form/OpeningPeriodForm.tsx
minevala: I'll figure something out

src/opening-period/form/OpeningPeriodForm.tsx
minevala: I'll add a tests. I may have to extract these

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,322 @@
...
+  apiTimeSpans: TimeSpanApiFormat[]
+): TimeSpanFormFormat[] {
+  return apiTimeSpans
+    .sort((a: TimeSpanApiFormat, b: TimeSpanApiFormat) => {
+      return getLowestWeekdayNumber(a) - getLowestWeekdayNumber(b);
+    })
+    .map((apiTimeSpan) => {
+      const weekdays: boolean[] = Array(7)
+        .fill(false)
```
VilleMiekkoja, line 85: Not sure if this `fill(false)` is necessary?

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,322 @@
...
+        endTime: apiTimeSpan.end_time
+          ? dropMilliseconds(apiTimeSpan.end_time)
+          : '',
+        resourceState: apiTimeSpan.resource_state || ResourceState.OPEN,
+        weekdays: weekdays as FormWeekdays,
+      };
+    });
+}
+
+const formatResourcePeriodOptionsToFormFormat = (
```
VilleMiekkoja, line 107: Should this be named __formatResourceStateOptionsToFormFormat__ instead?

src/opening-period/description/OpeningPeriodDescription.tsx
VilleMiekkoja: Not sure if it needs a separate component. To me it's good already like this, and abstracting it to a separate component would just increase the amount of jumps reader have to do, without actually removing any code really.

src/opening-period/form/OpeningPeriodForm.tsx
minevala: This was completely redundant type declaration.

src/opening-period/form/OpeningPeriodForm.tsx
minevala: This was completely redundant type declaration.

src/opening-period/form/OpeningPeriodForm.tsx
minevala: I'll rename

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+      return {
+        id: apiTimeSpan.id,
+        description: apiTimeSpan.description?.fi || '',
+        startTime: apiTimeSpan.start_time
+          ? dropMilliseconds(apiTimeSpan.start_time)
+          : '',
+        endTime: apiTimeSpan.end_time
+          ? dropMilliseconds(apiTimeSpan.end_time)
+          : '',
+        resourceState: apiTimeSpan.resource_state || ResourceState.OPEN,
```
VilleMiekkoja, line 94: Are we sure the default resource_state is open if there's no resource_state for the time span returned by api?

minevala: Nope, I'll remove this

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+  const { fields, append, remove } = useFieldArray({
+    control,
+    keyName: 'timeSpanUiId',
+    name: 'timeSpans',
+  });
+  const history = useHistory();
+  const [submitStatus, setSubmitStatus] = useState<SubmitStatus>('init');
+
+  const onSubmit = async (data: OpeningPeriodFormData): Promise<void> => {
+    const validTimeSpans: TimeSpanFormFormat[] = data.timeSpans.filter(
```
VilleMiekkoja, line 181: Is it possible for the user to input timeSpans that are not valid? Does this filter out those timeSpans that were added but where no field was inputted?

minevala: I do not think so. But to change this requires that we need to come up another way to initialise empty timespan

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+  DatePeriodOptions,
+  FormWeekdays,
+  Language,
+  ResourceState,
+  ResourceStateApiOption,
+  ResourceStateOption,
+  TimeSpan as TimeSpanApiFormat,
+  TimeSpanFormFormat,
+} from '../../common/lib/types';
+import { transformToApiFormat } from '../../common/utils/date-time/format';
```
VilleMiekkoja, line 15: It might be good idea to rename this to more specific, since it's applied outside of the file

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+  const resourceStateOptionsInApiFormat: ResourceStateApiOption[] =
+    datePeriodOptions.actions.POST.resource_state.choices;
+
+  const resourceStateOptions: ResourceStateOption[] = formatResourcePeriodOptionsToFormFormat(
+    resourceStateOptionsInApiFormat
+  );
+
+  const spans: TimeSpanFormFormat[] | {}[] =
+    datePeriod?.time_span_groups[0] &&
```
VilleMiekkoja, line 144: I wonder if we could/should show some kind of warning in the UI if API returns date period with more than one time span group? Though since we are adding the groups to the UI soon, it would be temporary and probably not worth it then.

src/opening-period/form/OpeningPeriodForm.tsx
minevala: This is much older function than the PR, but I can rename this

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,315 @@
...
+                  index={index}
+                  remove={remove}
+                />
+              )
+            )}
+          </div>
+          <SecondaryButton
+            dataTest="add-new-time-span-button"
+            onClick={(): void => append({})}
+            className="opening-period-final-action-button add-new-time-span-button">
```
VilleMiekkoja, line 234: This className `opening-period-final-action-button` might need some renaming. Originally it was only applied for the last final action buttons. But now (and actually for some time already) it's also applied in this '+ Lisää aukioloaika' button. And Im not sure if '+ Lisää aukiolaika' is fit for the 'final action button' 😅 . Not in the scope of this PR though, we can do it later.

minevala: Ok, we can figure out something else for this 👍  Or I can remove the class

src/opening-period/form/OpeningPeriodForm.tsx
VilleMiekkoja: Oh good point. It seems it was applied in EditOpeningPeriodPage before, but not in CreateNewOpeningPeriodPage. Not directly in the scope of this PR, we can do it later

src/opening-period/form/form-helpers/form-helpers.ts
```diff
@@ -0,0 +1,67 @@
+import {
```
VilleMiekkoja, line 1: Good call to abstract this! 👍

src/common/lib/types.ts
minevala: Fixed

src/common/utils/date-time/format.ts
minevala: Added tests

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Removed declaration

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Removed declaration

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Renamed as submitSuccessNotificationText etc.

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Extracted and added tests

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Renamed

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Removed default value

src/opening-period/form/OpeningPeriodForm.tsx
minevala: I renamed the function as transformDateToApiFormat

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Removed the class

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -1,189 +1,58 @@
...
-    return (
-      <>
-        <h1 className="resource-info-title">Aukiolojakson lisäys</h1>
-        <p>Sivua ladataan...</p>
-      </>
-    );
-  }
-
-  if (hasLoadingResourceError) {
+  if (hasDataLoadingError) {
```
VilleMiekkoja, line 55: Good naming change! 👍

src/opening-period/page/CreateNewOpeningPeriodPage.tsx
```diff
@@ -198,99 +67,40 @@ export default function CreateNewOpeningPeriodPage({
...
-            Julkaise
-          </PrimaryButton>
-          <SecondaryButton
-            className="add-new-opening-period-final-action-button"
-            onClick={(): void => history.push(`/resource/${resourceId}`)}>
-            Peruuta ja palaa
-          </SecondaryButton>
-        </div>
-      </form>
+      {resource && datePeriodOptions && (
```
VilleMiekkoja, line 87: I think these should already exist if the app gets this far. But is it because of typescript that we have to do this check?

minevala: Partly yes, partly because of state-hooks 😁

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  const { register, handleSubmit, errors, control, setValue } = useForm<
+    OpeningPeriodFormData
+  >({
+    mode: 'all',
+    defaultValues: formValues,
+  });
+  const { fields, append, remove } = useFieldArray({
+    control,
+    keyName: 'timeSpanUiId',
```
VilleMiekkoja, line 113: Is this required? Didn't seem to find it from code hmm..

src/opening-period/form/OpeningPeriodForm.tsx
minevala: I tested this and validation will catch invalid timespans

src/opening-period/form/OpeningPeriodForm.tsx
minevala: React-hook logged a warning because it uses id as it’s own internal prop. It  conflicted with timespan’s own id (in edit mode)

VilleMiekkoja:
Would EditOpeningPeriodPage benefit from some unit tests? Couple corner cases it could test are:

1. How the page behaves when api returns status code > 200 for fetching the initial date period
2. The page is able to render correctly different kinds of date periods. In e2e test we only test for one kind of date periods.

minevala:
> Would EditOpeningPeriodPage benefit from some unit tests? Couple corner cases it could test are:
> 
> 1. How the page behaves when api returns status code > 2xx for fetching the initial date period
> 2. The page is able to render correctly different kinds of date periods. In e2e test we only test for one kind of date period.

I would implement those in separate PR

VilleMiekkoja:
> > Would EditOpeningPeriodPage benefit from some unit tests? Couple corner cases it could test are:
> > 
> > 1. How the page behaves when api returns status code > 2xx for fetching the initial date period
> > 2. The page is able to render correctly different kinds of date periods. In e2e test we only test for one kind of date period.
> 
> I would implement those in separate PR like we did with the create-form

Sounds a good plan!

src/opening-period/time-span/TimeSpan.tsx
```diff
@@ -35,55 +41,70 @@ export default function TimeSpan({
...
+        <div className="time-span-state-container">
+          <Select
+            id={`time-span-state-id-${index}`}
+            onChange={(selection: { label: string; value: string }): void => {
+              setValue(`timeSpans[${index}].resourceState`, selection.value);
+            }}
+            className="time-span-state"
+            defaultValue={resourceStateOptions.find(
+              ({ value }) =>
+                value === ((item.resourceState as unknown) as string)
```
VilleMiekkoja, line 88: Why do we first cast it to unknown and then to string?

src/opening-period/time-span/Weekdays.tsx
```diff
@@ -1,29 +1,34 @@
...
 import './Weekdays.scss';
+import { ArrayField } from 'react-hook-form';
+import { TimeSpanFormFormat, FormWeekdays } from '../../common/lib/types';
 
 const DayCheckbox = ({
   register,
   dataTest,
   timeSpanIndex,
   dayIndex,
+  weekdays = [],
```
VilleMiekkoja, line 11: I wonder if we could replace this with defaultChecked value? Im not sure if the component needs the whole weekdays array?

src/opening-period/time-span/TimeSpan.tsx
minevala: This the way to supress enimmäkseen to a string. We do this also elsewhere

src/opening-period/time-span/Weekdays.tsx
minevala: It did not work when I tested it the last time. I may have not casted it then but I do not see the problem since dayIndex is passed also

VilleMiekkoja:
Was there suppose to be a spinner when user clicks 'Julkaise'? I tested and it does not seem to appear:
image

VilleMiekkoja:
Think the margin-up is missing here:

src/opening-period/time-span/Weekdays.tsx
VilleMiekkoja: Ouh I see. If this is the way it works then let's leave it like it is.

minevala: I can also pass the value

minevala:
> Was there suppose to be a spinner when user clicks 'Julkaise'? I tested and it does not seem to appear:
> image

I'll put it back

VilleMiekkoja:
This required a lot of effort. Good job on handling it so well! 👍

VilleMiekkoja (approved):

PR #79, opened 2020-12-17
Hauki-271 Add tests for EditOpeningPeriodPage

minevala:
- Supress expected console.errors in CreateOpeningPeriodPage-tests
- Fix time-span description default value issue -- api value was ignored
- Use `<ul>`-element for time-span listing 
- Add tests for EditOpeningPage

### DIFF
--- a/src/opening-period/page/EditOpeningPeriodPage.test.tsx
+++ b/src/opening-period/page/EditOpeningPeriodPage.test.tsx
@@ -300,7 +300,7 @@ describe(`<EditNewOpeningPeriodPage />`, () => {
   });
 
   it('should show error notification when loading date period fails', async () => {
-    const error: Error = new Error('Failed to load date period form options');
+    const error: Error = new Error('Failed to load date period');
     const errorConsoleSpy = jest
       .spyOn(global.console, 'error')
       .mockImplementationOnce((cb) => cb);
--- a/src/opening-period/page/EditOpeningPeriodPage.test.tsx
+++ b/src/opening-period/page/EditOpeningPeriodPage.test.tsx
@@ -191,41 +191,65 @@ describe(`<EditNewOpeningPeriodPage />`, () => {
     });
 
     await act(async () => {
+      const firstTimeSpan = container.querySelector(
+        '[data-test="time-span-list"] > li:first-child'
+      );
+
+      if (!firstTimeSpan) {
+        throw new Error(
+          'Something went wrong in first time-span rendering of EditOpeningPeriodPage'
+        );
+      }
+
       expect(
-        container.querySelector('[data-test="weekdays-monday-0-checkbox"]')
+        firstTimeSpan.querySelector(
+          '[type="checkbox"][data-test^="weekdays-monday"]'
+        )
       ).toBeChecked();
 
       expect(
-        container.querySelector('[data-test="weekdays-tuesday-0-checkbox"]')
+        firstTimeSpan.querySelector(
+          '[type="checkbox"][data-test^="weekdays-tuesday"]'
+        )
       ).toBeChecked();
 
       expect(
-        container.querySelector('[data-test="weekdays-wednesday-0-checkbox"]')
+        firstTimeSpan.querySelector(
+          '[type="checkbox"][data-test^="weekdays-wednesday"]'
+        )
       ).toBeChecked();
 
       expect(
-        container.querySelector('[data-test="weekdays-thursday-0-checkbox"]')
+        firstTimeSpan.querySelector(
+          '[type="checkbox"][data-test^="weekdays-thursday"]'
+        )
       ).toBeChecked();
 
       expect(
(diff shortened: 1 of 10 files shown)


src/opening-period/page/EditOpeningPeriodPage.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+      expect(container.querySelector('#openingPeriodBeginDate')).toHaveValue(
+        '18.12.2020'
+      );
+      expect(container.querySelector('#openingPeriodEndDate')).toHaveValue(
+        '18.01.2021'
+      );
+    });
+  });
+
+  it('should render first the time-span when it has the weekdays in the beginning of the week', async () => {
```
VilleMiekkoja, line 181: Not sure if this could be more clear like this: 'Should render first the time-span which has the weekdays in the beginning of the week'?

src/opening-period/page/EditOpeningPeriodPage.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+      if (!container) {
+        throw new Error(
+          'Something went wrong in rendering of CreateNewOpeningPeriodPage'
+        );
+      }
+    });
+
+    await act(async () => {
+      expect(
+        container.querySelector('[data-test="weekdays-monday-0-checkbox"]')
```
VilleMiekkoja, line 198: How do we know that this element is first? Is it because of the index value 0? Im wondering if there could be some other way to verify, that those elements are before other elements in dom?

VilleMiekkoja:
Good job! 🏅

src/opening-period/page/EditOpeningPeriodPage.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+    citizen_url: 'kansalaisen puolen url',
+    admin_url: 'admin puolen url',
+  },
+};
+
+const testDatePeriod: DatePeriod = {
+  id: 1,
+  created: '2020-11-20',
+  modified: '2020-11-20',
+  is_removed: false,
```
VilleMiekkoja, line 67: I can't decide if it's a good idea to add attributes to these that we do not use in the app yet? On the other hand it serves as a good documentation what the object returned by api actually looks. But on the other hand the reader might think that everything he sees here has some purpose and usage, and he is exposed to a complexity he would not need to when he reads these.

src/opening-period/page/EditOpeningPeriodPage.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+    await act(async () => {
+      expect(await screen.getByRole('heading', { level: 1 })).toHaveTextContent(
+        'Aukiolojakson muokkaus'
+      );
+      expect(await screen.getByText('Sivua ladataan...')).toBeInTheDocument();
+    });
+  });
+
+  it('should show error notification when loading date period fails', async () => {
+    const error: Error = new Error('Failed to load date period form options');
```
VilleMiekkoja, line 306: Small improvement suggestion: I believe the error message here is little misleading, probably just due to forgotten copy-paste?

src/opening-period/page/EditOpeningPeriodPage.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+    let container: HTMLElement;
+
+    await act(async () => {
+      container = render(
+        <EditOpeningPeriodPage resourceId="tprek:8100" datePeriodId="1" />
+      ).container;
+
+      if (!container) {
+        throw new Error(
+          'Something went wrong in rendering of CreateNewOpeningPeriodPage'
```
VilleMiekkoja, line 157: CreateNewOpeningPeriodPage -> EditOpeningPeriodPage

src/opening-period/page/EditOpeningPeriodPage.test.tsx
minevala: Fixed

src/opening-period/page/EditOpeningPeriodPage.test.tsx
minevala: I changed the list to use the <ul> and <li> elements. Now the first or last <li> is used in assertions as a root element.

src/opening-period/page/EditOpeningPeriodPage.test.tsx
minevala: Removed unused values

src/opening-period/page/EditOpeningPeriodPage.test.tsx
minevala: Fixed

src/opening-period/page/EditOpeningPeriodPage.test.tsx
minevala: Fixed

VilleMiekkoja (approved):

PR #80, opened 2020-12-18
Hauki 254 page for resource not found

VilleMiekkoja:
image

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -14,7 +14,7 @@ body {
   overflow-y: overlay;
   display: flex;
   flex-direction: column;
-  font-family: HelsinkiGrotesk-Regular, Arial, sans-serif;
+  font-family: var(--hauki-default-font);
   margin: 0;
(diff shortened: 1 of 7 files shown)


src/index.scss
```diff
@@ -14,7 +14,7 @@ body {
   overflow-y: overlay;
   display: flex;
   flex-direction: column;
-  font-family: var(--font-default);
+  font-family: HelsinkiGrotesk-Regular, Arial, sans-serif;
```
minevala, line 17: Nice catch 👍 
Could this be in the variables as app-body-text or something?

minevala (commented):
Nice job 👍 
Only one tiny comment

src/index.scss
VilleMiekkoja: - [ ] Put body font-family to its own variable 'app-body-text'

minevala (approved):
Nice job 👍

PR #85, opened 2020-12-29
Hauki 280 opening type edit fix

minevala:
- Fix edit date-period bug where all the time-spans' unchanged type data is lost on save. Our form library, React-form-hook, could not register initial values (every value has to be registered) because HDS Dropdown-component does not support ref-attribute. Instead, we need to use the React-hook-form's Controller wrapper component to circumvent the problem: https://react-hook-form.com/get-started/#IntegratingwithUIlibraries
- Pass timespans' ids and the timespangroup id and period-id on save 

- Use PATCH-method when updating date-period to prevent data inconsistency issues

### DIFF
--- a/src/common/lib/types.ts
+++ b/src/common/lib/types.ts
@@ -57,7 +57,7 @@ export type FormWeekdays = [
 ];
 
 export type TimeSpanFormFormat = {
-  id?: number;
+  id?: string;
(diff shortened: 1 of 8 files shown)


mikkokeskinen (approved):
Looks okay to me

PR #88, opened 2021-01-05
Hauki 197 opening period rules feature

minevala:
- Add rule-field to add, edit and delete rules. 
- Unify timespan- and rule-lists' styles

Default state when there is no rules, for example when user is creating a date-period:
Näyttökuva 2021-1-5 kello 16 26 15

One rule:
Näyttökuva 2021-1-5 kello 16 27 11

Multiple rules:
Näyttökuva 2021-1-5 kello 16 28 20


Rikuoja (approved):
Looks excellent! I'm afraid I have no idea on the implementation details tho 😅

VilleMiekkoja:
I wonder if it would be a good idea to tell the user that there's a default rule, and that adding a rule is not compulsory? Maybe there was a separate jira task for this already? Atm when I use the form Im not sure if I should add the rule or not? The term 'Voimassaolosääntö' feels pretty compulsory without additional note about it.

VilleMiekkoja:
When I start creating a new opening period, there's default value for it's type:
image

Maybe this is an existing but, not related to this pr.

VilleMiekkoja:
Upon changing the unit of time, the layout seems to make a little jump:

https://user-images.githubusercontent.com/2777633/103995569-55713880-51a1-11eb-8dc1-3cb724de17ec.mov

VilleMiekkoja:
If I press 'add rule' in the form, then should all or some of the fields of the rule then become compulsory? It feels like I can add a rule, but select any of its fields, and the UI still allows user to post the form?

minevala:
> Maybe this is an existing bug, not related to this pr.

I can check this one indenpendently

minevala:
> Upon changing the unit of time, the layout seems to make a little jump:

I could set a fixed min-width to the last boxes so they would not jump so much

minevala:
> If I press 'add rule' in the form, then should all or some of the fields of the rule then become compulsory? It feels like I can add a rule, but not select any of its fields, and the UI still allows user to post the form?

We could try to hide these more complicated settings under "advanced" or something

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -63,9 +69,14 @@ export default function OpeningPeriodForm({
       ? formatApiTimeSpansToFormFormat(firstTimeSpanGroup.time_spans)
       : [{}];
 
+  const rules: GroupRule[] | [] = firstTimeSpanGroup?.rules.length
```
VilleMiekkoja, line 72: While reading this I was confused why rules come only from the first time span group, but then I realised we only have one time span group at moment.

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -76,26 +87,51 @@ export default function OpeningPeriodForm({
...
 
-  const { register, handleSubmit, errors, control } = useForm<
+  const { register, handleSubmit, errors, control, setValue } = useForm<
     OpeningPeriodFormData
   >({
     mode: 'all',
     defaultValues: formValues,
   });
+
   const { fields, append, remove } = useFieldArray({
```
VilleMiekkoja, line 100: I wonder if it would be good idea to rename these too for them to be consistent with how we name these in rows 107 - 109?

minevala: I can change these as well

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Yes, we still use only one group 😁

src/opening-period/page/CreateNewOpeningPeriodPage.test.tsx
```diff
@@ -171,6 +171,19 @@ describe(`<CreateNewOpeningPeriodPage />`, () => {
           },
         ],
       },
+      timeSpanGroup: {
```
VilleMiekkoja, line 174: Should we add a unit test for submitting opening period with (multiple?) rules also? The rendering of those rules seems to be tested in the editOpeningPeriodPage.test.tsx, but I couldn't find a unit test for submitting the form with time span rules? I wonder if it would be good to have, though at the same time I know unit testing this form has  been hard, as we have had problems with timeouts in them?

minevala: I could actually extend the current test 🤔

src/opening-period/page/EditOpeningPeriodPage.test.tsx
```diff
@@ -309,6 +372,84 @@ describe(`<EditNewOpeningPeriodPage />`, () => {
     });
   });
 
+  it('should render correct time-span-group rules', async () => {
```
VilleMiekkoja, line 375: Should we also test that user is able to add and delete time-span-group rules?

minevala: Maybe it should be done in the e2e-test?

minevala: In a way, add is already tested in the e2e-test for new date-period

src/opening-period/rule/Rule.tsx
```diff
@@ -0,0 +1,303 @@
...
+      frequencyOrdinalField,
+      selectedFrequency?.value.frequency_ordinal || null
+    );
+  };
+
+  useEffect(() => {
+    register({ name: frequencyModifierField });
+    register({ name: frequencyOrdinalField });
+
+    // "If you choose to not use Controller and manually register fields, you will need to update the input value with setValue." - https://react-hook-form.com/api#register
```
VilleMiekkoja, line 181: True, but I think this only applies in situations where Ref is not available, if I understood the documentation correct?

minevala: Yes, and we cannot use ref since frequency is derived of two api values: modifier and ordinal

minevala: It seems that react-hook-form does not initialize the value without using ref. At least I did not find that from the source code: https://github.com/react-hook-form/react-hook-form/blob/master/src/useForm.ts#L908

src/opening-period/form/OpeningPeriodForm.tsx
minevala: - [x] changed

src/opening-period/page/CreateNewOpeningPeriodPage.test.tsx
minevala: I think simulating user interactions makes testing hard. It would be nicer to test this by populating the fields and them sent them back to server. I guess this test already covers it for edit.
But if we want to test it for save also it requires more work.

src/opening-period/page/EditOpeningPeriodPage.test.tsx
minevala: - [x] Add delete rule test

src/opening-period/rule/Rule.tsx
```diff
@@ -0,0 +1,303 @@
...
+              options={contextOptions}
+              defaultValue={contextOptions.find(
+                (option: InputOption): boolean => option.value === value
+              )}
+              label="Valitse aika"
+              placeholder="Aikaväli"
+            />
+          )}
+        />
+        <Select
```
VilleMiekkoja, line 237: Why is this select not wrapped inside Controller, but the other Select components are?

src/opening-period/page/EditOpeningPeriodPage.test.tsx
VilleMiekkoja: It feels kind of a corner case, and I've usually preferred to test those in unit tests, in order to not increase the amount of e2e tests because I've felt them more expensive with potential to random errors etc.. But Im not sure, you can decide

VilleMiekkoja:
Good Job!

src/opening-period/rule/Rule.tsx
minevala: It is an input that is derivated from two values. It requires that we need to register two values, and it is possible with control. Maybe I could extract it into a component because of its different behavior.

VilleMiekkoja: Thanks for clarification. Probably no need to extract it imo, it's better inlined

VilleMiekkoja (approved):

VilleMiekkoja (approved):

VilleMiekkoja (approved):

PR #94, opened 2021-01-12
Openshift compatibility

Sukriva:
For better Openshift compatibility changed Nginx to Red Hat supported Nginx and set correct permissions for random non-root user. Also removed installing bash since the red hat supported nginx already has bash.


minevala:
Would it make sense to have two docker files side by side for a moment?
We could have one for the old environment (dockerfile_old) and the other for this new OpenShift environment (dockerfile). We could run both builds side by side since we are missing some build steps in the new environment.

Sukriva:
I checked with our architect and he said yes.

PR #97, opened 2021-01-14
HAUKI-293 Show listing of possible parent resources on a resource page

VilleMiekkoja:
image

### DIFF
--- a/src/resource/page/ResourcePage.scss
+++ b/src/resource/page/ResourcePage.scss
@@ -39,6 +39,6 @@
   margin-top: 0;
 }
 
-  .child-and-parent-resource-list-item {
+.child-and-parent-resource-list-item {
   margin-top: $spacing-m;
 }
--- a/src/resource/page/ResourcePage.scss
+++ b/src/resource/page/ResourcePage.scss
@@ -35,10 +35,10 @@
   white-space: pre-wrap;
 }
 
-.child-resource-description-text {
+.child-and-parent-resource-description-text {
   margin-top: 0;
 }
 
-.child-resource-list-item {
+  .child-and-parent-resource-list-item {
   margin-top: $spacing-m;
 }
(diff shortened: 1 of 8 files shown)


src/resource/page/ResourcePage.scss
```diff
@@ -35,10 +35,10 @@
   white-space: pre-wrap;
 }
 
-.child-resource-description-text {
+.child-and-parent-resource-description-text {
```
VilleMiekkoja, line 38: Not sure if Im satisfied by this class name 🤔 . What do you think? It's a class for both child and parent resource description. Would it be better if it was: 'child-or-parent-resource-description-text'? One option would be just 'resource-description-text'. But I feel it's not enough explicit in what it means. Because the whole page is about the resource itself, and it includes the child and parent resources

src/resource/page/ResourcePage.tsx
```diff
@@ -142,12 +143,22 @@ export default function ResourcePage({ id }: { id: string }): JSX.Element {
...
-          setChildResources(childR);
+        const resourceHasChildren = r.children.length > 0;
+        const resourceHasParents = r.parents.length > 0;
+
+        if (resourceHasChildren || resourceHasParents) {
+          // fetch children and parents
+          const [childsResponse, parentsResponse] = await Promise.all([
+            resourceHasChildren
+              ? api.getChildResourcesByParentId(r.id)
+              : { results: [] },
```
minevala, line 154: I think this might be a bit simpler if the api-method would return a list?

minevala (commented):
Nice 👍

src/resource/page/ResourcePage.scss
VilleMiekkoja: - [ ] Rename to related-resource-

src/resource/page/ResourcePage.tsx
VilleMiekkoja: - [ ] Return list from api-method

minevala (approved):
Nice job 👍

PR #98, opened 2021-01-15
Hauki 233 render timespan groups

minevala:
- Update opening-period form implementation so it can show multiple timegroups
- Tweak styles of opening-period-form's timespans and rules

localhost_3000_resource_34299

### DIFF
--- a/cypress/integration/AddNewOpeningPeriod.spec.ts
+++ b/cypress/integration/AddNewOpeningPeriod.spec.ts
@@ -33,21 +33,21 @@ describe('User adds a new opening period', () => {
     cy.get('button.dayButton').contains('01').click();
 
     // Then enter time span data
-    cy.get('[data-test=weekdays-monday-0]').click();
-    cy.get('[data-test=weekdays-tuesday-0]').click();
-    cy.get('[data-test=weekdays-wednesday-0]').click();
-    cy.get('[data-test=weekdays-thursday-0]').click();
-    cy.get('[data-test=weekdays-friday-0]').click();
-    cy.get('[data-test=time-span-start-time-0]').type('08:00');
-    cy.get('[data-test=time-span-end-time-0]').type('16:00');
-    cy.selectHdsDropdown({ id: 'time-span-state-id-0', value: 'Auki' });
+    cy.get('[data-test=weekdays-monday-0-0]').click();
+    cy.get('[data-test=weekdays-tuesday-0-0]').click();
+    cy.get('[data-test=weekdays-wednesday-0-0]').click();
+    cy.get('[data-test=weekdays-thursday-0-0]').click();
+    cy.get('[data-test=weekdays-friday-0-0]').click();
+    cy.get('[data-test=time-span-start-time-0-0]').type('08:00');
+    cy.get('[data-test=time-span-end-time-0-0]').type('16:00');
+    cy.selectHdsDropdown({ id: 'time-span-state-id-0-0', value: 'Auki' });
 
     // Enter rules data
-    cy.get('button[data-test="add-new-rule-button"]').click();
-    cy.selectHdsDropdown({ id: 'rule-0-context', value: 'Kuukausi' });
-    cy.selectHdsDropdown({ id: 'rule-0-frequency', value: 'Parillinen' });
-    cy.selectHdsDropdown({ id: 'rule-0-subject', value: 'Viikko' });
-    cy.selectHdsDropdown({ id: 'rule-0-start', value: '1.' });
+    cy.get('button[data-test="add-new-rule-button-0"]').click();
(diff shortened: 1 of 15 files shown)


cypress/integration/AddNewOpeningPeriod.spec.ts
```diff
@@ -33,21 +33,21 @@ describe('User adds a new opening period', () => {
...
+    cy.get('[data-test=time-span-end-time-0-0]').type('16:00');
+    cy.selectHdsDropdown({ id: 'time-span-state-id-0-0', value: 'Auki' });
 
     // Enter rules data
-    cy.get('button[data-test="add-new-rule-button"]').click();
-    cy.selectHdsDropdown({ id: 'rule-0-context', value: 'Kuukausi' });
-    cy.selectHdsDropdown({ id: 'rule-0-frequency', value: 'Parillinen' });
-    cy.selectHdsDropdown({ id: 'rule-0-subject', value: 'Viikko' });
-    cy.selectHdsDropdown({ id: 'rule-0-start', value: '1.' });
+    cy.get('button[data-test="add-new-rule-button-0"]').click();
```
VilleMiekkoja, line 46: Adding rules data seems like a corner-case, since 99% of users do not need to add this rule. I suggest we would test corner-cases only in unit tests, since on average they seem to be more reliable

minevala: I can remove this test if it is unnecessary

VilleMiekkoja:
Should we add unit tests for user creating/editing multiple timespan groups?

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -221,76 +184,46 @@ export default function OpeningPeriodForm({
...
+              defaultValue={`${timeSpanGroup.id || ''}`}
+              ref={register()}
+            />
+            <input
+              type="hidden"
+              name={`${timeSpanGroupFieldName}[${index}].period`}
+              defaultValue={`${timeSpanGroup.period || ''}`}
+              ref={register()}
+            />
+            <TimeSpans
```
VilleMiekkoja, line 209: It was a good idea to extract TimeSpans and Rules to their own component 👍

minevala:
> Should we add unit tests for user creating/editing multiple timespan groups?
Those unit tests files are pretty long already, and I noticed that it has become hard to fix failing tests. I prefer e2e-tests over unit tests for core functionality. But if rules are not part of the core functionality, we should figure out something else. Maybe we should add the first tests to form itself.

VilleMiekkoja:
> > Should we add unit tests for user creating/editing multiple timespan groups?
> 
> Those unit tests files are pretty long already, and I noticed that it has become hard to fix failing tests. I prefer e2e-tests over unit tests for the core functionality. But if rules are not part of the core functionality, we should figure out something else. Maybe we should add the first tests to form itself. What do you think, @VilleMiekkoja ?

I have some thoughts on the matter:

I've previously preferred e2e testing for the most common happy case, and then all other cases including cornercases would be tested either by unit tests or integration tests. This is because I've had bad experiences in testing corner-cases with e2e. What happened was:

1. Testing cornercases with e2e increased the time of test runs by a lot. Even now it seems our e2e tests in CI take way longer than unit tests, even we have much more unit tests than e2e tests I believe
image

2. With more e2e tests we had more random errors, as unit tests seemed to be more robust. This seems to be the case with us also; In my experience the unit tests have had way less random errors than e2e tests. If we increase the amount of e2e tests, we might see an increase in the random errors.

But tbf, we have had our issues with unit tests as well. Writing them has been more cumbersome, that's for sure. And there has been some stability issues with them as well, even though I would say way less than in e2e tests. Though then again the stability issues of e2e tests can reveal real bugs. It's a shame that we too have almost adopted the 'retry' culture here instead of trying to more actively solve the actual problem behind the random errors. Im definitely guilty of that too, it's just easier in the moment to simply retry instead of taking the time to investigate the real problem behind.

Anyway I think it would be nice if this is consistent. So if we decide to test cornercases with e2e, then it would be good to test all the cornercases there instead of unit tests. Also then we would have to think again what role does unit test serve in our project to begin with? We would have to implement all the cornercases we currently test with unit tests to e2e tests. That would take some time.

Another solution is to try to improve our unit tests. If the files are getting bigger, we could extract stuff to multiple files instead. I think our application at hand is very complex, so it is expected that the tests reflects that complexity. Testing the complexity can result in large amount of test code, which we have to manage some way. Though again tbf, there is more boilerplate in unit tests than there is in e2e tests. So I do understand that there would probably not be as much code if we would test it in e2e test.

What do you think?

minevala:
My idea was to add first OpeningPeriodForm "unit" tests for the form itself:
- Should render time-span-group's rules
- Add to the time-span-group
- Edit the rule in the time-span-group

VilleMiekkoja (approved):

PR #99, opened 2021-01-20
Clear errors from test logs

minevala:


src/opening-period/page/CreateNewOpeningPeriodPage.test.tsx
```diff
@@ -10,6 +10,13 @@ import {
 import api from '../../common/utils/api/api';
 import CreateNewOpeningPeriodPage from './CreateNewOpeningPeriodPage';
 
+// Opening period form uses react-router to redirect after successful save - mocking prevents unnecessary errors in test logs
+jest.mock('react-router-dom', () => ({
+  useHistory: (): { push: jest.Mock } => ({
+    push: jest.fn(),
```
VilleMiekkoja, line 16: Should we check that the push function is called in the tests? I was thinking that would be a way to test that the redirect really happens.

minevala: It is already checked in the e2e-tests. But we could check it here instead 🤔

VilleMiekkoja (approved):

PR #100, opened 2021-01-20
HAUKI-55 Show resource opening hours

VilleMiekkoja:
image


src/resource/resource-opening-hours/opening-period/OpeningPeriod.tsx
```diff
@@ -109,6 +202,47 @@ export default function OpeningPeriod({
           confirmText="Poista"
         />
       </div>
+      {isOpen && (
+        <div className="date-period-details-container">
```
minevala, line 206: It might be easier to read and test this if this would be an independent component. I would also extract the resolving of nonsupported - logic inside that component.

VilleMiekkoja: Good idea:

- [ ] Extract date period details into its own component

src/common/utils/date-time/format.ts
```diff
@@ -28,3 +29,113 @@ export const formatDateRange = ({
...
+  return weekdays[language][weekdayIndex];
+}
+
+type WeekdaySpan = {
+  startIndex: number;
+  lastInsertedIndex: number;
+  endIndex?: number;
+};
+
+export function createWeekdaysStringFromIndices(
```
VilleMiekkoja, line 87: This is very likely not the most elegant way to solve the creation of the weekdays string :D. If you figure and elegant alternative I'll change this.

Rikuoja (approved):
Looks good!

minevala (approved):
Nice 👍

PR #102, opened 2021-01-25
Hauki 233 upsert time span group

minevala:
- Add add-timespan-group button and helper text 
- Add remove-time-span-group button (has a bug still, form will preserve old values when trying to append a new group after removing the old)
- Add shared test-utils

Näyttökuva 2021-1-25 kello 11 55 02


Rikuoja (approved):
This clarifies the UI for the user, looks good!

src/opening-period/form/OpeningPeriodForm.test.tsx
```diff
@@ -48,15 +57,14 @@ const testDatePeriodOptions: UiDatePeriodConfig = {
...
-const groupIdB = 2;
+const groupIdA = 100;
+const groupIdB = 200;
 const timeSpanGroupA = {
   id: groupIdA,
   period: testDatePeriodId,
   rules: [
     {
       id: 1,
-      group: groupIdA,
```
VilleMiekkoja, line 59: Why are these deleted? Are the groupIds applied somewhere else?

minevala: I do not remember anymore but I think we should send them back 🤔

minevala: I fixed it so that the group-id will be passed as form-data. Every new rule or timespan will get current group id if it has one.

VilleMiekkoja:
Well done with this one! 🏅

VilleMiekkoja (approved):

PR #104, opened 2021-01-28
Hauki 315 update time group section styles

minevala:
Style changes according to new design: 
- Update form-component texts
- Add icons to add-rule-button and add-time-span-button
- Update other styles of time-span-group's form-components

Näyttökuva 2021-1-28 kello 16 59 06


VilleMiekkoja (approved):
Looking good! 👍

PR #105, opened 2021-01-28
HAUKI-313 Fix page crashing when start or end time was not specified for time span

VilleMiekkoja:
image


src/resource/resource-opening-hours/opening-period/PeriodOpeningHours.test.tsx
```diff
@@ -95,6 +95,93 @@ const testDatePeriod: DatePeriod = {
   ],
 };
 
+const testDatePeriodWithMissingStartAndEndTimes: DatePeriod = {
```
minevala, line 98: It would generate less code if this would just be a modified version of the testDatePeriod:
```js
const testDatePeriodWithMissingStartAndEndTimes: DatePeriod = {
     ...testDatePeriod,
     timeSpanGroups: [
         { ...timeSpanGroups[0], timeSpans: [{...}, {...}] }
     ]
}
```

src/resource/resource-opening-hours/opening-period/PeriodOpeningHours.test.tsx
```diff
@@ -95,6 +95,93 @@ const testDatePeriod: DatePeriod = {
...
+            fi: null,
+            sv: null,
+            en: null,
+          },
+          description: {
+            fi: 'Aukiolon testikuvaus',
+            sv: null,
+            en: null,
+          },
+          start_time: '',
```
minevala, line 128: Does API return also empty strings?

src/resource/resource-opening-hours/opening-period/PeriodOpeningHours.test.tsx
VilleMiekkoja: - [ ] Reduce the amount of test code by applying the previously created testDatePeriod

src/resource/resource-opening-hours/opening-period/PeriodOpeningHours.test.tsx
VilleMiekkoja: Actually it probably does not, so I'll change this to null:

- [ ] Change empty string to null

minevala: Ok, I thought there was something, hopefully we are right 😬 😁

minevala (approved):
Nice work 👍

PR #109, opened 2021-02-03
Hauki 369 add validations and restrictions to date-period basic data

minevala:
- Add max-length restriction to date-period name field (use api value for it)
- Validate date-range on submit

Näyttökuva 2021-2-3 kello 10 02 02

### DIFF
--- a/src/opening-period/description/OpeningPeriodDescription.tsx
+++ b/src/opening-period/description/OpeningPeriodDescription.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { FieldErrors } from 'react-hook-form/dist/types/errors.d';
 import { TextArea, TextInput } from 'hds-react';
 import './OpeningPeriodDescription.scss';
+import { TextFieldConfig } from '../../common/lib/types';
 
 type TFieldValues = {
   openingPeriodTitle: string;
@@ -10,12 +11,16 @@ type TFieldValues = {
 export default function OpeningPeriodDescription({
   register,
   errors,
+  nameFieldConfig,
 }: {
   register: Function;
   errors: FieldErrors<TFieldValues>;
+  nameFieldConfig: TextFieldConfig;
 }): JSX.Element {
   const hasError =
     errors.openingPeriodTitle && errors.openingPeriodTitle.type === 'required';
+
+  const titleMaxLength: number | undefined = nameFieldConfig.max_length;
 
   return (
     <>
@@ -31,11 +36,15 @@ export default function OpeningPeriodDescription({
           data-test="openingPeriodTitle"
           id="openingPeriodTitle"
           aria-invalid={errors.openingPeriodTitle ? 'true' : 'false'}
-          ref={register({ required: true, maxLength: 100 })}
+          ref={register({
+            required: true,
+            maxLength: titleMaxLength,
+          })}
           helperText={
             hasError ? 'Aukiolojakson otsikko on pakollinen' : undefined
           }
           invalid={hasError}
+          {...(titleMaxLength ? { maxLength: titleMaxLength } : {})}
         />
       </div>
       <div className="form-control">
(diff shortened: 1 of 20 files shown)


src/opening-period/description/OpeningPeriodDescription.tsx
```diff
@@ -31,11 +36,15 @@ export default function OpeningPeriodDescription({
...
-          ref={register({ required: true, maxLength: 100 })}
+          ref={register({
+            required: true,
+            maxLength: titleMaxLength,
+          })}
           helperText={
             hasError ? 'Aukiolojakson otsikko on pakollinen' : undefined
           }
           invalid={hasError}
+          {...(titleMaxLength ? { maxLength: titleMaxLength } : {})}
```
VilleMiekkoja, line 47: Does maxLength prop from here go to the native input of html as 'maxlength'? I thought react maps camelCase to kebab-case. Since it seems the correct form of html input maxlength is written like 'maxlength', without '-'.

minevala: Yes, the HDS component passes it to native html input

src/opening-period/form/OpeningPeriodForm.test.tsx
```diff
@@ -173,7 +176,67 @@ const renderOpeningPeriodForm = (props: OpeningPeriodFormProps): Element => {
 };
 
 describe(`<OpeningPeriodForm />`, () => {
-  describe(`TimeSpanGroups`, () => {
+  describe('Validations', () => {
```
VilleMiekkoja, line 179: Good looking tests 👍

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -54,6 +58,20 @@ export type OpeningPeriodFormProps = {
   errorTextAndLabel: NotificationTexts;
 };
 
+const validateStartInputWithEndDate = (end: Date | null) => (
+  start: string | null
+): boolean | string => {
```
VilleMiekkoja, line 63: Is this how the react-hook-form validation works? So that if you return a boolean value true, it is valid, but if you return anything else than exactly that (for example a text string), it interpretes it as invalid?

minevala: Yes, this is how it seems to work:
- true === valid
- (false || string) === invalid

Here is an example code of custom validation: https://github.com/react-hook-form/react-hook-form/blob/master/examples/customValidation.tsx

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -183,7 +206,11 @@ export default function OpeningPeriodForm({
               labelText="Alkaa"
               onChange={(value): void => setPeriodBeginDate(value || null)}
               value={periodBeginDate}
+              error={errors.openingPeriodBeginDate}
               registerFn={register}
+              customValidations={{
```
VilleMiekkoja, line 211: I wonder would it be better to have the validation in the end time input field, as the end time is often inserted last, and it is the field where user was/is when the error is prompted?

minevala: Yes. It makes sense, I'll fix

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -193,8 +217,21 @@ export default function OpeningPeriodForm({
               onChange={(value): void => setPeriodEndDate(value || null)}
               value={periodEndDate}
               registerFn={register}
+              customValidations={{
+                dateRange: validateEndInputWithStartDate(periodBeginDate),
```
VilleMiekkoja, line 221: Tested the feature locally and it looked nice! 👍  I wonder if we could/should trigger the validation after user has selected the dates? Currently the validation is triggered by user trying to send the form. I guess it could be ok to begin with, and then we could improve it later on.

minevala: I think it is doable with relative small changes. I’ll test 👍

VilleMiekkoja (approved):
Well done! 🏅

PR #111, opened 2021-02-03
Added Azure DevOps pipeline

Sukriva:
This pipeline extends the base pipeline and defines CI and PR triggers in GitHub

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,18 +11,6 @@ trigger:
     exclude:
     - README.md
 
-# Pull request (PR) triggers cause a pipeline to run whenever a pull request is 
-# opened with one of the specified target branches, or when updates are made to 
-# such a pull request.
-#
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,38 @@
...
+    - README.md
+
+# Pull request (PR) triggers cause a pipeline to run whenever a pull request is 
+# opened with one of the specified target branches, or when updates are made to 
+# such a pull request.
+#
+# GitHub creates a new ref when a pull request is created. The ref points to a 
+# merge commit, which is the merged code between the source and target branches 
+# of the pull request.
+pr:
```
VilleMiekkoja, line 21: After discussing with Jussi Arpalahti we decided the pull-request pipeline would be a separate from this. So according to current plan we would not need this pipeline to be triggered from pull-requests and therefore this could be removed?

Sukriva:
Closing this for now and revisiting this after other pipeline changes are done.

PR #116, opened 2021-02-08
Hauki 370 add rules and timespans validations

minevala:
- Timespan has to have either start or end time:
Näyttökuva 2021-2-8 kello 10 49 05

- Rule's mandatory fields are resolved from API OPTIONS response: 
Näyttökuva 2021-2-8 kello 10 48 57

+ React-hook-form lib updated
+ Opening-period form uses now FormContext (no need pass the most common methods downwards as props)


src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -184,138 +198,131 @@ export default function OpeningPeriodForm({
...
-        <SecondaryButton
-          className="opening-period-final-action-button"
-          onClick={(): void => returnToResourcePage()}>
-          Peruuta ja palaa
-        </SecondaryButton>
-      </div>
-    </form>
+              />
+            </section>
+            {errors.openingPeriodEndDate?.type === 'dateRange' &&
```
VilleMiekkoja, line 240: Are beginDate errors checked also?

VilleMiekkoja:
This might be a dumb question but in rule validation, why are frequency and rule start optional fields? Do they default to some values?

src/opening-period/form/OpeningPeriodForm.tsx
minevala: No, because we decided to go with the end date validation 🤔 
It has changes because it uses common ErrorText-component

VilleMiekkoja:
Should weekday fields be also compulsory at this stage? I mean at the moment it does not create anything sensible if you leave them out? Or does it implicitly mean every weekday is selected if none is selected? Not sure if that is clear to the users. Maybe we should have a toggle button there to selected/deselect them all

VilleMiekkoja:
Date period time span (alkamispäivä ja loppumispäivä) validation seems to be broken. Think I forgot to do a test for it initially. The form let's you submit itself without selecting begin and/or end date. But api responds with an error.

minevala:
> Date period time span (alkamispäivä ja loppumispäivä) validation seems to be broken. Think I forgot to do a test for it initially. The form let's you submit itself without selecting begin and/or end date. But api responds with an error.

How this can be reproduced?

VilleMiekkoja (approved):
Well done 👍

PR #117, opened 2021-02-08
HAUKI-372 Fix creation of exception period

VilleMiekkoja:
Creating an exception period resulted as a regular period before. This pull-request fixes the bug.

### DIFF
--- a/src/opening-period/form/OpeningPeriodForm.tsx
+++ b/src/opening-period/form/OpeningPeriodForm.tsx
@@ -52,7 +52,7 @@ type NotificationTexts = {
 export type OpeningPeriodFormProps = {
   formId: string;
   datePeriod?: DatePeriod;
-  forceException: boolean;
+  forceException?: boolean;
   resourceId: number;
   datePeriodConfig: UiDatePeriodConfig;
   submitFn: (datePeriod: DatePeriod) => Promise<DatePeriod>;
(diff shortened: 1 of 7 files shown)


src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -47,6 +47,7 @@ type NotificationTexts = {
 export type OpeningPeriodFormProps = {
   formId: string;
   datePeriod?: DatePeriod;
+  forceException: boolean;
```
minevala, line 50: Could this be optional attribute just for CreatePeriodPage? Something like initialExceptionData or initialFormData and `{ override: true }`?
Then it would not be necessary to use the attribute in the EditPeriodPage

VilleMiekkoja: This attribute is not required by EditPeriodPage. It is only applied in CreateNewOpeningPeriodPage and OpeningPeriodForm.

But maybe your point is that it's confusing to pass that as a prop for OpeningPeriod form inside the EditPeriodPage? Kind of wouldn't want an abstraction object for only one simple boolean value though.

VilleMiekkoja: - [ ] Make forceException optional and to default false

minevala (approved):
👍

PR #120, opened 2021-02-11
Hauki 378 improve rule frequency field

minevala:
- Add default value for rule's frequency "Jokainen"
- Disable rule's start field if 'Pariton' or 'Parillinen' frequency is selected

Näyttökuva 2021-2-11 kello 14 50 37

### DIFF
--- a/src/opening-period/page/EditOpeningPeriodPage.test.tsx
+++ b/src/opening-period/page/EditOpeningPeriodPage.test.tsx
@@ -648,17 +648,7 @@ describe(`<EditNewOpeningPeriodPage />`, () => {
             {
               id: 1225,
               period: 1,
-              rules: [
-                {
-                  context: 'month',
-                  frequency_modifier: null,
-                  frequency_ordinal: 1,
-                  id: 20,
-                  group: 1225,
-                  subject: 'week',
-                  start: 2,
-                },
-              ],
+              rules: [testDatePeriod.time_span_groups[0].rules[1]],
               time_spans: testDatePeriod.time_span_groups[0].time_spans,
             },
           ],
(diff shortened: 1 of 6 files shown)


src/opening-period/page/EditOpeningPeriodPage.test.tsx
```diff
@@ -687,4 +658,76 @@ describe(`<EditNewOpeningPeriodPage />`, () => {
       expect(mockHistoryPush).toHaveBeenCalledWith(testResourcePageUrl);
     });
   });
+
+  it('should disable rule start when "Parillinen" frequency is selected', async () => {
```
VilleMiekkoja, line 662: It seems we have multiple places where we could have this test, and I don't really have a strong opinion which place would be best:

1. EditNewOpeningPeriodPage
2. CreateNewOpeningPeriodPage
3. OpeningPeriodForm
4. Rule.test.tsx

What do you think which place should have which tests?

src/opening-period/rule/Rule.tsx
```diff
@@ -151,51 +152,50 @@ export default function Rule({
...
-      selectedFrequency?.value.frequency_modifier || null
-    );
+    setValue(frequencyModifierField, selectedModifier || null);
     setValue(
       frequencyOrdinalField,
       selectedFrequency?.value.frequency_ordinal || null
     );
-  };
+
+    if (selectedModifier) {
```
VilleMiekkoja, line 185: Does this mean it is either 'pariton' or 'parillinen'?

src/opening-period/page/EditOpeningPeriodPage.test.tsx
minevala: I could move rule test under OpeningPeriodForm or Rule.test.tsx? 🤔

src/opening-period/rule/Rule.tsx
minevala: Yes, start does not apply anything to the rule when it has a modifier. I could add a constant which would make easier to read tough 🤔

minevala: I added a comment instead :grin

VilleMiekkoja (approved):
Good job 🎖️

PR #122, opened 2021-02-12
HAUKI-380 Add option to select full day when selecting begin and end time of a time span

VilleMiekkoja:
Unchecked:
image

Checked:
image

On smaller screen:
image


src/opening-period/time-span/TimeSpan.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const methods = useForm({ defaultValues });
+    return <FormProvider {...methods}>{children}</FormProvider>;
+  };
+
+  return {
+    ...render(component, { wrapper: Wrapper }),
+  };
+}
+
+describe(`<TimeSpan />`, () => {
```
minevala, line 23: Should all the timespan tests locate in the same place?

src/opening-period/time-span/TimeSpan.tsx
```diff
@@ -88,18 +90,23 @@ export default function TimeSpan({
               dataTest={`time-span-start-time-${groupIndex}-${index}`}
               registerFn={(ref): void =>
                 register(ref, {
-                  validate: (startTime): boolean =>
-                    validateTimeRange(
+                  validate: (startTime): boolean => {
+                    if (fullDay) {
```
minevala, line 94: This condition could probably locate inside the validateTimeRange-function. Then the name of the function should be something else tough 🤔

src/opening-period/time-span/TimeSpan.scss
```diff
@@ -37,3 +38,13 @@
...
 }
+
+.checkbox-container {
+  display: flex;
+  align-items: center;
+  height: var(--spacing-3-xl);
+}
+
+.time-span-end-time-input-wrapper {
+  margin-right: 140px;
```
minevala, line 49: Could this be replaced with some predefined design token? So there would be some rhythm in the layout: '$spacing-layout-2-xl'

src/opening-period/time-span/TimeSpan.scss
```diff
@@ -37,3 +38,13 @@
   justify-content: space-between;
   align-items: flex-end;
 }
+
+.checkbox-container {
+  display: flex;
+  align-items: center;
+  height: var(--spacing-3-xl);
```
minevala, line 45: Could this be in pixels with a comment? // 56px is the input height etc.

minevala (commented):
Works nicely 👍

src/opening-period/time-span/TimeSpan.test.tsx
VilleMiekkoja: This is a good question. I think they all can't be in this file, since some of the validations require sending the form. And sending is not available if only TimeSpan component is rendered. So maybe validation tests should be with the full form, and other features could be tested by individual components?

minevala: There might still be some tests that could be here. I think we need some conventions for organizing tests.
I'm worried about test overlap; tests are also liability 🤔

minevala: Maybe it could be something like this:
- rendering tests (including conditional rendering, ui changes) -> component tests
- save tests, validation tests -> form tests
- page rendering -> page tests

src/opening-period/time-span/TimeSpan.scss
VilleMiekkoja: - [ ] Add comment about pixel value and reasoning

src/opening-period/time-span/TimeSpan.scss
VilleMiekkoja: - [ ] Apply design token here instead of hard coded value

minevala (approved):
Splendid 😁 👍

PR #126, opened 2021-02-16
Fix useArrayField types

minevala:
Fix useArrayField types:
1. Initialize arrayField type in the useArrayField call by providing the value type and the key for ui-id field (uses 'id' by default):

      React-hook-form type implementation for useArrayFields is like this: 
      ```TFieldArrayValues extends Record<string, any> = Record<string, any>, TKeyName extends string = "id"```
      
      ...so we can provide:
      `useArrayField<TimeSpanFormFormat, 'timeSpanUiId'>` 
2. Add types to fields map callback using ArrayField and providing the previous type as value type. 
Needs to be partial because the fields are typed as partial in the react-hook-form's type implementation:
      ```ts
        fields.map(timeSpanGroup: Partial<ArrayField<TimeSpanFormFormat, 'timeSpanUiId'>> => )
      ```

### DIFF
--- a/src/opening-period/form/OpeningPeriodForm.tsx
+++ b/src/opening-period/form/OpeningPeriodForm.tsx
@@ -11,7 +11,6 @@ import {
   DatePeriod,
   UiDatePeriodConfig,
   Language,
-  TimeSpanFormFormat,
   TimeSpanGroupFormFormat,
   UiFieldConfig,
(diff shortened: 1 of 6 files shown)


VilleMiekkoja (approved):
DIdn't understand but this seems to simplify existing code so I'll approve it, nice work 👍

PR #129, opened 2021-02-19
Hauki 387 fixed action footer to form

VilleMiekkoja:
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -50,84 +50,112 @@ export default function App(): JSX.Element {
     <div className="App">
       <AuthContext.Provider value={{ authTokens, clearAuth }}>
         <Router>
-          <NavigationAndFooterWrapper>
-            <Main id="main">
-              <Switch>
-                <Route exact path="/">
+          <Switch>
+            <Route exact path="/">
+              <NavigationAndFooterWrapper renderFooter>
+                <Main id="main">
                   <h1>Etusivu</h1>
-                </Route>
-                <Route exact path="/not_found">
+                </Main>
+              </NavigationAndFooterWrapper>
+            </Route>
+            <Route exact path="/not_found">
+              <NavigationAndFooterWrapper renderFooter>
+                <Main id="main">
                   <h1>Kohdetta ei löydy</h1>
                   <p>
                     Kohdetta ei löytynyt. Yritä myöhemmin uudestaan. Ongelman
                     toistuessa ota yhteys sivuston ylläpitoon. Teidät on
                     automaattisesti kirjattu ulos.
                   </p>
-                </Route>
-                <Route exact path="/unauthorized">
+                </Main>
+              </NavigationAndFooterWrapper>
+            </Route>
+            <Route exact path="/unauthorized">
+              <NavigationAndFooterWrapper renderFooter>
+                <Main id="main">
                   <h1>Puutteelliset tunnukset</h1>
-                </Route>
-                <Route exact path="/unauthenticated">
+                </Main>
+              </NavigationAndFooterWrapper>
+            </Route>
+            <Route exact path="/unauthenticated">
+              <NavigationAndFooterWrapper renderFooter>
+                <Main id="main">
                   <h1>Puuttuvat tunnukset</h1>
-                </Route>
-                <PrivateResourceRoute
-                  id="resource-route"
-                  exact
-                  path="/resource/:id"
-                  render={({
-                    match,
-                  }: RouteComponentProps<{ id: string }>): ReactElement => (
+                </Main>
+              </NavigationAndFooterWrapper>
+            </Route>
+            <PrivateResourceRoute
+              id="resource-route"
+              exact
+              path="/resource/:id"
+              render={({
+                match,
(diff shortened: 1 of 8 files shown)


src/opening-period/form/OpeningPeriodForm.scss
```diff
@@ -84,7 +85,21 @@
 }
 
 .opening-period-final-action-row-container {
-  margin-top: 116px;
+  height: $spacing-layout-xl;
+  position: fixed;
```
minevala, line 89: Could it actually be a sticky action row?

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -354,17 +354,19 @@ export default function OpeningPeriodForm({
           </p>
         </div>
         <div className="opening-period-final-action-row-container">
```
minevala, line 356: className="opening-period-form-action-row"?

src/opening-period/form/OpeningPeriodForm.scss
```diff
@@ -35,6 +35,7 @@
 
 .form-actions-row-condensed {
   padding: 0 $spacing-l;
+  margin-bottom: $spacing-layout-2-xl;
```
minevala, line 38: Could this be a variable based on the height of the floating action row? I assume with sticky position this is not even needed

src/App.tsx
```diff
@@ -50,84 +50,112 @@ export default function App(): JSX.Element {
     <div className="App">
       <AuthContext.Provider value={{ authTokens, clearAuth }}>
         <Router>
-          <NavigationAndFooterWrapper>
-            <Main id="main">
-              <Switch>
-                <Route exact path="/">
+          <Switch>
+            <Route exact path="/">
+              <NavigationAndFooterWrapper renderFooter>
```
minevala, line 55: `<NavigationAndFooterWrapper renderFooter>` -> `<PageWithNavigation withFooter>`?

minevala: Hmm, now when I rethink, my suggestion is not the the greatest. But I was thinking about the name now when the footer is optional.

src/opening-period/form/OpeningPeriodForm.scss
VilleMiekkoja: With sticky I have problem of getting it 100% width of the page 🤔

src/opening-period/form/OpeningPeriodForm.tsx
VilleMiekkoja: There's already a class named form-actions-row, so it would probably semantically overlap with it 🤔

src/App.tsx
VilleMiekkoja: True it is little confusing, since it won't always render the Footer. Would it be better to just render Navigation only for the form pages? Think I tried something like that before during this story but there was some problem with it. I could try it again?

minevala: Yes, maybe it would more clear with that one 🤔

minevala: But if it gets too complicated, we can go with the current solution too 👍

src/opening-period/form/OpeningPeriodForm.scss
```diff
@@ -84,7 +85,21 @@
 }
 
 .opening-period-final-action-row-container {
-  margin-top: 116px;
+  height: $spacing-layout-xl;
```
minevala, line 88: It is missing the box-shadow

src/opening-period/form/OpeningPeriodForm.scss
VilleMiekkoja: - [ ] Apply final action row container height

src/opening-period/form/OpeningPeriodForm.scss
VilleMiekkoja: - [ ] Insert missing box-shadow

src/App.tsx
VilleMiekkoja: - [ ] Render only Navigation component to form pages instead of NavigationAndFooter wrapper

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -354,17 +354,19 @@ export default function OpeningPeriodForm({
...
-            className="opening-period-final-action-button"
-            type="submit">
-            Julkaise
-          </PrimaryButton>
-          <SecondaryButton
-            className="opening-period-final-action-button"
-            onClick={(): void => returnToResourcePage()}>
-            Peruuta ja palaa
-          </SecondaryButton>
+          <div className="main-container">
```
minevala, line 357: I think this div could be a exported component in the Main.tsx so we will not accidentally delete or modify it.

minevala: I think it is pretty much ready then

VilleMiekkoja: - [ ] Create and apply MainContainer component

minevala (approved):
Nice 👍

PR #130, opened 2021-02-19
Hauki 328 date period status

minevala:
**Date period has resource-state set:**
Näyttökuva 2021-2-22 kello 13 26 52

**Date period has no resource-state set:**
Näyttökuva 2021-2-19 kello 16 57 01

### DIFF
--- a/src/opening-period/form/OpeningPeriodForm.tsx
+++ b/src/opening-period/form/OpeningPeriodForm.tsx
@@ -11,6 +11,7 @@ import {
   DatePeriod,
   Language,
   LanguageStrings,
+  ResourceState,
   TimeSpanGroupFormFormat,
   UiFieldConfig,
   UiFormRuleConfig,
@@ -28,6 +29,7 @@ import {
   SupplementaryButton,
 } from '../../components/button/Button';
 import { ErrorText } from '../../components/icon-text/IconText';
+import ResourceStateSelect from '../../components/resourse-state-select/ResourceStateSelect';
 import toast from '../../components/notification/Toast';
 import {
   formatTimeSpanGroupsToApiFormat,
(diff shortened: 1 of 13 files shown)


src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -255,90 +274,134 @@ export default function OpeningPeriodForm({
...
+        </div>
+        {isResourceStateSet(resourceStateValue) ? (
+          <section
+            key="resource-state-notification"
+            data-test="resource-state-notification"
+            className="form-section time-span-group">
+            <Notification
+              label="Aukiolojaksolle on valittu aukiolotyyppi"
+              className="opening-period-form-notification">
+              Kun aukiolojaksolle on valittu aukiolotyyppi niin se poistaa
```
minevala, line 350: "niin se poistaa _tämän_"

minevala: - [x] Fixed

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -173,8 +182,15 @@ export default function OpeningPeriodForm({
...
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     getValues,
+    watch,
   } = formMethods;
 
+  const clearResourceState = (): void => {
+    setValue(openingPeriodResourceStateKey, 'undefined');
+  };
+
+  const resourceStateValue = watch(openingPeriodResourceStateKey);
```
VilleMiekkoja, line 192: Does watch work so that it creates a re-render of this component when the value associated with the openingPeriodResourceStateKey changes? That way then for example on line 327 can re-render itself? So with this watch we can avoid the need of applying useState in our components and can apply the watch instead, atleast in some cases?

VilleMiekkoja (approved):
Nice 👍

src/opening-period/form/OpeningPeriodForm.tsx
minevala: Yes, it is specially meant for that.

PR #136, opened 2021-02-26
Use correct date-format to parse initial API dates

minevala:
**Bug**: validation failed when the user tried to edit any field value of date-period with the same start-date and end-date.
**Cause**: initial start-date string is converted to date with JavaScript's Date class without date-format. The API will return the date in the 'yyyy-MM-dd' - format. The JavaScript Date will resolve the current time based on GMT threshold, so it will be 02:00. 
Date-fns will use 00:00 time when time is not defined in the date-format.
**Fix**: parse the API dates with date-fns parse function.


VilleMiekkoja (approved):
Good catch! 👍

PR #140, opened 2021-03-26
created new pipeline scripts

jaana-ibm:
Created two new Azure pipeline scripts to match the new pipeline solution. At the same time deleted the old azure-pipeline.yml file.


VilleMiekkoja (approved):
Good job 🏅

PR #144, opened 2021-04-21
Hauki 439 fix e2e env variable

minevala:
The e2e-tests read the resource id from the HAUKI_RESOURCE env variable. It makes e2e-tests more portable.
Here is the Cypress doc: https://docs.cypress.io/guides/guides/environment-variables#Option-5-Plugins


Rikuoja (approved):

VilleMiekkoja (approved):
Good job! 🎉

PR #146, opened 2021-04-26
Feature/hds datepicker

minevala:
- Upgrade HDS libraries to 1.1.0.
- Use new HDS DateInput component. Fixes the issue when the user typed the input did not register the first keypress and its value.
- Use onBlur validation in the form to prevent unnecessary validation cycles during typing 

New dateinput:
Näyttökuva 2021-5-3 kello 15 32 24
Näyttökuva 2021-5-4 kello 9 52 57
Näyttökuva 2021-5-4 kello 9 53 03


VilleMiekkoja (approved):
Well done! 👍

PR #151, opened 2021-05-26
Feature/hds timeinput

minevala:
- Use HDS TimeInput component in opening hours inputs to improve usability
Näyttökuva 2021-5-28 kello 16 52 28


src/opening-period/time-span/TimeSpan.tsx
```diff
@@ -32,6 +31,26 @@ const descriptionPlaceholderTexts: LanguageStrings = {
   en: 'Esim. for women only',
 };
 
+const validateTimeFormat = (timeString: string): boolean => {
+  const [hours, minutes] = timeString.split(':');
+  return (
+    (!hours && !minutes) ||
```
VilleMiekkoja, line 37: Does this mean that the time format is valid if there's no hours and no minutes?

VilleMiekkoja:
Good work 👍

src/opening-period/time-span/TimeSpan.tsx
minevala: I could rename this or something. I agree, the name does not make sense 👋

VilleMiekkoja (approved):
Nice 👍

PR #173, opened 2021-12-01
Adjust the most volatile e2e-tests' timeouts

minevala:
The e2e-tests of the application fail quite often. It is not always the same tests that fail everytime but it is always some of the certain group of tests. Usually the tests fail because API does not response within the timeout limits (4000ms - 5000ms). It takes often more than 5000ms from API to respond one or two consecutive requests. 
Now the timeout limits are raised in the most problematic places up to 10000ms.


VilleMiekkoja (approved):
Good change, let's hope it helps 👍

PR #184, opened 2022-01-03
Use react-hook-form's Controller for correct value in the fullday checkbox

minevala:
### Changes
- Use Controller component to wrap HDS Checkbox component (controlled component)
- With Controller component we can ensure that the value of the fullday Checkbox is a boolean and the react-hook-form is using correct name for the value
- React-hook-form [Controlled component](https://react-hook-form.com/api/usecontroller/controller)

### Fixes
- User was not able to add a full day time span to an opening period which already had a timespan


VilleMiekkoja (approved):
Good fix!

PR #189, opened 2022-02-07
Feature/closing button

minevala:
### Description
- Refactor internal Button component to use HDS theme constant or theme variables
- Add iconRight, light, and size props for internal Secondary button
- Use internal Secondary button in opening-period list headers to ensure similar styles
- Add AppContext for hasOpenerWindow value and closeAppWindow method
- Add conditional close-button (also signs out user) into App header
- Replace user menu with user name text 

### Motivation
- An service user can close the window when ready

Screenshot 2022-02-08 at 16 40 37


VilleMiekkoja (approved):
Good job! 🎉

PR #196, opened 2022-04-22
Update React, React Scripts & React Hook Form

pirkkahuhtala:
Initially this was about updating React Hook Form to the latest version v7. Because of that we needed to update React and React Scripts as well. Because this change there are some new linting warnings appearing. We need to fix those at some point.


src/common/utils/api/api.test.ts
```diff
@@ -15,7 +15,7 @@ describe('apiRequest', () => {
   });
 
   describe('request', () => {
-    it('adds auth-tokens into every request', async (done) => {
+    it('adds auth-tokens into every request', async () => {
```
pirkkahuhtala, line 18: We should use either async or done. It started to complain about it.

src/opening-period/form/OpeningPeriodForm.tsx
```diff
@@ -278,46 +282,53 @@ export default function OpeningPeriodForm({
             <h3 className="opening-period-section-title">Ajanjakso</h3>
             <div className="form-control">
               <section className="opening-period-time-period">
-                <DateInput
-                  id="openingPeriodBeginDate"
+                <Controller
```
pirkkahuhtala, line 285: The DateInput onChange was incompatible with the one returned from the register.

pirkkahuhtala: Probably could have done some sort of mapping in the register.

pirkkahuhtala:
Resolved here: https://github.com/City-of-Helsinki/hauki-admin-ui/pull/230

PR #197, opened 2022-04-25
Add initial Hauki load tests

pirkkahuhtala:
Adds initial load tests for the project. The load tests are handled with [k6](https://k6.io) which needs to be installed in your system. I added instructions to the README.

To run the load tests you need to execute this command:
```
$ ./k6/run.sh hauki-scenarios.js
```

### DIFF
--- a/k6/README.md
+++ b/k6/README.md
@@ -1,0 +1,15 @@
+# Hauki load testing
+
+Scripts for load testing Hauki API.
+
+## Prerequisities
+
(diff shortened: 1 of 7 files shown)


k6/README.md
```diff
@@ -0,0 +1,21 @@
+# Hauki load testing
```
minevala, line 1: Would it be easier to understand the purpose of the folder if the name was api-perf-tests etc. instead k6?

PR #203, opened 2022-06-14
Remove unneeded content from the front page

pirkkahuhtala:
Begin transforming the front page to apply to the new designs.


minevala (approved):
🍕

PR #206, opened 2022-06-15
Use preview for period page

pirkkahuhtala:


src/common/utils/api/api.ts
```diff
@@ -275,7 +277,14 @@ export default {
...
       resourceState: {
-        options: resourceStateOptions,
+        options: [
+          ...resourceStateOptions,
+          // TODO: This needs to be returned from the server
+          {
+            label: { fi: 'Muu, mikä?', sv: 'Muu, mikä', en: 'Muu, mikä?' },
+            value: ResourceState.OTHER,
+          },
+        ],
```
minevala, line 287: The API data model does not support this kind of free-form state. And probably never will because the state has to be an enumeration.
I think we need to handle this in the client if the state is open and the opening has a title then it is other.
What we can do later on is to add more common enumerations to API.

src/constants.ts
```diff
@@ -0,0 +1,26 @@
+import { TranslatedApiChoice } from './common/lib/types';
+import { Rule } from './opening-period/v2/types';
+
+// eslint-disable-next-line import/prefer-default-export
+export const uiRules: TranslatedApiChoice<Rule>[] = [
+  {
+    value: 'week_every',
```
minevala, line 7: Should we have these enumerations in the common/types.td?

src/components/opening-hours-rows/OpeningHoursRows.tsx
```diff
@@ -0,0 +1,173 @@
+import React, { Fragment } from 'react';
```
minevala, line 1: Would opening-hours-preview describe more what this is?

minevala: And then the other one could be form-opening-hours-preview since it is a wrapper for this

src/resource/resource-opening-hours/opening-period/OpeningPeriod.tsx
```diff
@@ -68,9 +69,6 @@ export default function OpeningPeriod({
       className="opening-period"
       data-test={`openingPeriod-${datePeriod.id}`}>
       <div className="opening-period-header">
-        <div className="opening-period-dates opening-period-header-column">
-          <div>{formattedDateRange}</div>
-        </div>
         <div className="opening-period-title opening-period-header-column">
           {datePeriodName ? (
             <h4>{datePeriodName}</h4>
```
minevala, line 74: This title now seems too small when there are even and odd week titles: 
Screenshot 2022-06-15 at 14 20 30

minevala (commented):
Nice code reuse again 👍

src/resource/resource-opening-hours/opening-period/OpeningPeriod.tsx
pirkkahuhtala: Fixed

PR #209, opened 2022-06-16
Fix resource page button focus

pirkkahuhtala:


azure/pull-request.pipeline.yml
```diff
@@ -30,9 +30,9 @@ steps:
     CI=true yarn test
   displayName: 'Yarn run tests'
 
-# - script: |
-#    CI=true yarn test-cypress-with-new-test-api
-#   displayName: 'Yarn run cypress (e2e) tests'
+- script: |
```
minevala, line 33: These will probably fail at the moment. Should we leave them in comments for now?

src/components/language-select/LanguageSelect.scss
```diff
@@ -1,5 +1,20 @@
...
+  position: relative;
+}
+
+.custom-language-select:focus-within {
+  box-shadow: none;
+}
+
+.custom-language-select:focus-within::after {
+  border: 2px solid var(--color-white);
+  bottom: -2px;
```
minevala, line 12: I would add "display: block" just to avoid lineheight issues and just for clarity.

minevala (commented):
Nice fix 👍 
I added a couple of suggestions.

src/components/language-select/LanguageSelect.scss
pirkkahuhtala: Ah yes. I'll fix it. Thanks. 👍

minevala (approved):
👍

PR #220, opened 2022-06-28
Add e2e for adding opening hours

pirkkahuhtala:
Add e2e test for the scenario where resource maintainer adds new opening hours.

Steps:
- Navigate to add new opening hours
- Set resource open from Monday till Friday 08.00-16.00
- Set Saturday from 10.00-15.00 as open and from 15.00-16.00 as self service
- Set Sunday as closed
- Save

### DIFF
--- a/src/components/opening-hours-weekdays/DayCheckbox.tsx
+++ b/src/components/opening-hours-weekdays/DayCheckbox.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Language } from '../../common/lib/types';
 import { getWeekdayShortNameByIndexAndLang } from '../../common/utils/date-time/format';
+import { nameToId } from '../../common/utils/form/form';
 import './DayCheckbox.scss';
 
 const DayCheckbox = ({
@@ -14,11 +15,12 @@ const DayCheckbox = ({
   onChange: (checked: boolean) => void;
   checked?: boolean;
 }): JSX.Element => {
-  const id = `${namePrefix}-weekdays-${currentDay}`;
+  const id = nameToId(`${namePrefix}-weekdays-${currentDay}`);
 
   return (
     <label htmlFor={id} className="day-label">
       <input
+        data-test={id}
         id={id}
         type="checkbox"
         onChange={(): void => {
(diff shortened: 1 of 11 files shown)


src/components/opening-hours-weekdays/DayCheckbox.tsx
```diff
@@ -14,11 +15,12 @@ const DayCheckbox = ({
   onChange: (checked: boolean) => void;
   checked?: boolean;
 }): JSX.Element => {
-  const id = `${namePrefix}-weekdays-${currentDay}`;
+  const id = sanitizeId(`${namePrefix}-weekdays-${currentDay}`);
```
minevala, line 18: Should the name be stringifyBrackets or something?
And it should be used like this in here:
```js
const id =`${stringifyBrackets(namePrefix)-weekdays-${currentDay}`;
```

Or could this be getUiId or something else? And then it could take parts like this:
`getUiId([namePrefix, 'weekdays', currentDay])`

minevala (commented):
Nice first e2e-test 🏅 
One little suggestion
I was thinking we could have a testing strategy where we have a few e2e-tests and bunch of unit tests. I would stay away from complex component or integration tests for now

src/components/opening-hours-weekdays/DayCheckbox.tsx
pirkkahuhtala: > Nice first e2e-test 🏅 One little suggestion I was thinking we could have a testing strategy where we have a few e2e-tests and bunch of unit tests. I would stay away from complex component or integration tests for now

That sounds good. Would it mean that we could remove the complex component/integration tests in this change?

minevala: > That sounds good. Would it mean that we could remove the complex component/integration tests in this change?
In this or in some other PR. Then we would have some tests and then it would be easier to change functionality if/when needed

pirkkahuhtala: Looks like we do not have that many of those anymore.

pirkkahuhtala: > Should the name be stringifyBrackets or something? And it should be used like this in here:
> 
> ```js
> const id =`${stringifyBrackets(namePrefix)-weekdays-${currentDay}`;
> ```
> 
> Or could this be getUiId or something else? And then it could take parts like this: `getUiId([namePrefix, 'weekdays', currentDay])`

Did the changes for the brackets removal. The `getUiId` feels brilliant. Was thinking that we could have another function which would return both name and id so we could get both in one call but did not yet go to that route..

minevala (approved):
🍻

PR #223, opened 2022-07-15
Feature/holidays forms

minevala:
## Description 
- Add holidays page for editing resource's single holidays in one place
- Added a few missing date period properties to OpeningHourFormValues
- Fix to opening-hour-helper to accept a Date Period without timeSpanGroups (Date Periods can be always open)
 
 ## Motivation
 - We do not know yet but it might be easier for the user to view yearly holidays on one page. Now we have something for the end users to test
 
 ## How is this tested?
 - Locally on the dev machine
 - unit tests


pirkkahuhtala (approved):
Sweet feature for adding the holidays. Good job!

PR #239, opened 2022-08-10
Refactor: reorganise the opening hours form code

pirkkahuhtala:
### Description
Re-organise the OpeningHoursForm component code

### Motivation
Make the code a bit more cleaner. Maybe in the future we could put these actions in its` own hook.

### How is this tested?
- Locally on the dev machine
- e2e tests

### DIFF
--- a/src/components/opening-hours-form/OpeningHoursForm.tsx
+++ b/src/components/opening-hours-form/OpeningHoursForm.tsx
@@ -8,7 +8,6 @@ import {
   Resource,
   ResourceState,
   UiDatePeriodConfig,
-  OpeningHours as TOpeningHours,
   OpeningHoursFormValues,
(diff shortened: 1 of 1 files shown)


src/components/opening-hours-form/OpeningHoursForm.tsx
```diff
@@ -154,11 +154,44 @@ const OpeningHoursForm = ({
       timeSpanGroups: [defaultTimeSpanGroup],
     };
     insert(newIdx, values, { shouldFocus: false });
-    // FIXME: For some reason the normal array won't get added in the insert
-    setValue(`openingHours.${newIdx}`, values);
```
pirkkahuhtala, line 158: Now that we have the v7 from the hook form library we could get rid of this.

